PR #3, opened 2022-07-22
added zIndex to popover element

BrancuAlexandru:


components/Nav/Nav.tsx
```diff
@@ -13,7 +13,7 @@ const navigation = [
 
 const Nav: NextPage = () => {
   return (
-    <Popover as="header" className="relative">
+    <Popover as="header" className="relative" style={{zIndex: 10}}>
```
pkspyder007, line 16: use tailwindcss classes for z-index

NiallJoeMaher: To second Praveen - [https://tailwindcss.com/docs/z-index](https://tailwindcss.com/docs/z-index)

z-10 should be enough

NiallJoeMaher (changes requested):
Could you remove the `package-lock` and `yarn.lock`?

They shouldn't be changed unless there are package changes and made thousands of line changes. 
Screenshot 2022-07-23 at 09 45 23
`

PR #26, opened 2022-07-30
feat: switch from Date to Temporal

DogPawHat:
https://tc39.es/proposal-temporal/docs/

ran into some cases where the server hydration would break because of a difference in locale between server and client. Proper fix is to restrict it to client side only, but for now I'll hardcode it to `en-IE`, and ditch the useless Date for the glorious if still in Stage 3 Temporal.

### DIFF
--- a/components/ArticlePreview/ArticlePreview.tsx
+++ b/components/ArticlePreview/ArticlePreview.tsx
@@ -1,5 +1,6 @@
 import type { NextPage } from "next";
 import Link from "next/link";
+import { Temporal } from '@js-temporal/polyfill';
 
 type Props = {
   title: string;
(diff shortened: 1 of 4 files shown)


NiallJoeMaher:
This is very cool! I didn't even know it was an issue. I need to do some reading on this Temporal stuff. 🤔

NiallJoeMaher:
> https://tc39.es/proposal-temporal/docs/


This is very cool! I didn't even know it was an issue. I need to do some reading on this Temporal stuff. 🤔

DogPawHat:
It's the future of time itself in JS. Currently it's blocked by the IETF needing to standardize an extension to RFC3339/ISO8601https://github.com/tc39/proposal-temporal/issues/1450

NiallJoeMaher (approved):
🌮 Cool thing to learn about! Cheers dude

PR #27, opened 2022-08-01
feat: use react markdown in place of inner html

DogPawHat:
Proposal to replace `markdown-it` with `react-markdown` component (biased on remark and rehype), mostly so we don't have to use `dangerouslySetInnerHtml`. (See https://github.com/remarkjs/react-markdown#security)


NiallJoeMaher (approved):
🌮 On fire, pretty cool! Never came across markdownIt

Will make it easier when we add our own dashboard for writing articles.🤘

PR #28, opened 2022-08-01
chore: replace img with next imgs in static pages

DogPawHat:
Lotta linter warnings about this, so gonna give this a shot.

### DIFF
--- a/components/ArticlePreview/ArticlePreview.tsx
+++ b/components/ArticlePreview/ArticlePreview.tsx
@@ -1,5 +1,6 @@
 import type { NextPage } from "next";
 import Link from "next/link";
+import Image from "next/image";
 
 type Props = {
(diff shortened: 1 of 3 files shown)


NiallJoeMaher:
I'm gonna test this branch! I know that dang Image tag can behave weirdly. That's why I didn't use it all straight away

NiallJoeMaher:
Yup as I suspected, this throws a few warnings/errors
Screenshot 2022-08-01 at 19 22 25

DogPawHat:
Ah ok, it's not happy if you just use tailwind classes then ☹️

NiallJoeMaher:
@DogPawHat yes! Let's close it unless we can figure out a way of making it work. Or we can revisit later. 🦾

PR #55, opened 2022-10-05
add codeowners file

Nikki9523:
-Add codeowners.md file
-Not sure if the double tab is needed but I saw it here so I added it, let me know if I should remove it https://github.blog/2017-07-06-introducing-code-owners


NiallJoeMaher (changes requested):
Thanks for this!

This one throws everyone, but the filename should be `CODEOWNERS` NOT `CODEOWNERS.md`.

Nikki9523:
Ah thanks @NiallJoeMaher :) I replaced the .md file with just a CODEOWNERS file

NiallJoeMaher (approved):
🌮

PR #89, opened 2022-10-28
Fixed padding issues with the author information.

VSB14:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
- The Padding attributes "px-4" have been removed from Biobar.tsx file to fix the alignment issues with the author information at the end of each article.
## Any Breaking changes:
- None
## Associated Screenshots:
    Before Commit
image

    After Commit
image

### DIFF
--- a/components/BioBar/BioBar.tsx
+++ b/components/BioBar/BioBar.tsx
@@ -15,8 +15,8 @@ const BioFooter = ({ author }: Props) => {
 
   const { name, image, bio, username } = author;
   return (
-    <div className="max-w-xl px-4 mx-auto  text-gray-700 dark:text-gray-300 mt-6">
-      <div className="flex mx-2 sm:mx-6 md:mx-auto px-4 border-t-2 pt-6 border-gray-300 dark:border-gray-800">
+    <div className="max-w-xl mx-auto  text-gray-700 dark:text-gray-300 mt-6">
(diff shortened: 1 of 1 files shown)


components/BioBar/BioBar.tsx
```diff
@@ -15,8 +15,8 @@ const BioFooter = ({ author }: Props) => {
 
   const { name, image, bio, username } = author;
   return (
-    <div className="max-w-xl px-4 mx-auto  text-gray-700 dark:text-gray-300 mt-6">
-      <div className="flex mx-2 sm:mx-6 md:mx-auto px-4 border-t-2 pt-6 border-gray-300 dark:border-gray-800">
+    <div className="max-w-xl mx-auto  text-gray-700 dark:text-gray-300 mt-6">
```
auxfuse, line 18: Could we consider removing the double-space present in this line between `mx-auto` & `text-gray-700` ? 🤔 It's not a big deal that won't break anything, but it is a nice standard to adhere to 😅

auxfuse (changes requested):
Nice change here! Just one thing for feedback considerations 🤜

pkspyder007:
@auxfuse I think we should add linting config and git hooks to automate these small issues.
can you open some issues for the same, It would be some nice beginners issues

components/BioBar/BioBar.tsx
NiallJoeMaher: Nice catch! Eagle eyes 🦅

PR #95, opened 2022-10-30
Fetches host URL from ServerSideProps

NiallJoeMaher:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
Change to grabbing the host URL from getServerSideProps so that the `og:url` will work in preview environments too. 
Also fixes a typo that had a copy-paste from niall.af

## Any Breaking changes:
None

## Associated Screenshots:
None

### DIFF
--- a/pages/articles/[slug].tsx
+++ b/pages/articles/[slug].tsx
@@ -13,9 +13,9 @@ import rehypePrism from "rehype-prism";
 
 const ArticlePage: NextPage = ({
   post,
+  host,
 }: InferGetServerSidePropsType<typeof getServerSideProps>) => {
   if (!post) return null;
-
   return (
     <>
       <Head>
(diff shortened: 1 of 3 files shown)


pages/articles/[slug].tsx
```diff
@@ -28,13 +28,10 @@ const ArticlePage: NextPage = ({
...
         <meta property="og:type" content="article" />
-        <meta
-          property="og:url"
-          content={`https://codu.co/articles/${post.slug}`}
-        />
+        <meta property="og:url" content={`${host}/articles/${post.slug}`} />
         <meta
           property="og:image"
-          content={`${process.env.BASE_URL || `http://localhost:3000`}/api/og?title=${post.title}`}
+          content={`${host}/api/og?title=${post.title}`}
```
Comment (jh3y) [inline pages/articles/[slug].tsx:34]: Love it! 💪 Wondering if there is some kind of win where this could even be defined in the `Layout` 🤔 in a blanket style using page title and an override where defined.

jh3y (approved):

pages/articles/[slug].tsx
Comment (NiallJoeMaher) [reply pages/articles/[slug].tsx:34]: Just wondering would we have to server side render everything then. 🤔 Or were you thinking something different?

pkspyder007 (approved):

PR #101, opened 2022-10-31
Allow save with empty canonical URL

### LINKED ISSUE #100: Fix bug when saving article if "View advanced settings" is open.
## Context

When trying to save an article, after viewing the advanced settings (which contains the Canonical URL) I am unable to save without a valid canonical URL.

## Expected Behavior

We should be able to save with an empty canonical URL.

## Current Behavior

We cannot save with an empty canonical URL if the "advanced settings" has been opened.

NiallJoeMaher:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
Updates the save schema to allow saving with an empty canonical URL which was causing a bug. 

Closes #100

### DIFF
--- a/schema/post.ts
+++ b/schema/post.ts
@@ -14,7 +14,7 @@ export const SavePostSchema = z.object({
     .trim()
     .max(156, "Max length is 156 characters.")
     .optional(),
-  canonicalUrl: z.string().trim().url().optional(),
+  canonicalUrl: z.string().trim().url().or(z.string().length(0)).optional(),
   tags: z.string().array().max(5).optional(),
 });
 


schema/post.ts
```diff
@@ -14,7 +14,7 @@ export const SavePostSchema = z.object({
     .trim()
     .max(156, "Max length is 156 characters.")
     .optional(),
-  canonicalUrl: z.string().trim().url().optional(),
+  canonicalUrl: z.string().trim().url().or(z.string().length(0)).optional(),
```
pkspyder007, line 17: Can't we use `.min(0)` instead of `.or(...) `

pkspyder007 (approved):

PR #124, opened 2022-12-31
Add ability to upload profile photo with s3

NiallJoeMaher:
# ✨ Codu Pull Request 💻

## Pull Request details:
- Adds Dockerfile to test build on AWS
- Removes use of  Nextjs Image when using bucket for profile photo until we have a domain attached 

## Any Breaking changes:
To upload photos locally you need to setup an s3 bucket

Screenshot 2022-12-31 at 16 17 21

### DIFF
--- a/server/trpc/router/profile.ts
+++ b/server/trpc/router/profile.ts
@@ -1,6 +1,14 @@
-import { saveSettingsSchema, getProfileSchema } from "../../../schema/profile";
+import {
+  saveSettingsSchema,
+  getProfileSchema,
+  uploadPhotoUrlSchema,
+  updateProfilePhotoUrlSchema,
+} from "../../../schema/profile";
+
+import { getPresignedUrl } from "../../common/getPresignedUrl";
 
 import { router, publicProcedure, protectedProcedure } from "../trpc";
+import { TRPCError } from "@trpc/server";
 
 export const profileRouter = router({
   edit: protectedProcedure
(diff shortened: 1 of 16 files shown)


server/trpc/router/profile.ts
```diff
@@ -16,6 +24,43 @@ export const profileRouter = router({
...
+      const extension = type.split("/")[1];
+
+      if (!["jpg", "jpeg", "gif", "png", "webp"].includes(extension)) {
+        throw new TRPCError({
+          code: "BAD_REQUEST",
+          message: "Invalid file format.",
+        });
+      }
+
+      if (size > 8388608) {
```
pkspyder007, line 53: can we keep this somewhere as a constant?

server/trpc/router/profile.ts
```diff
@@ -16,6 +24,43 @@ export const profileRouter = router({
...
+      });
+      return profile;
+    }),
+  getUploadUrl: protectedProcedure
+    .input(uploadPhotoUrlSchema)
+    .mutation(async ({ ctx, input }) => {
+      const { size, type } = input;
+      const extension = type.split("/")[1];
+
+      if (!["jpg", "jpeg", "gif", "png", "webp"].includes(extension)) {
```
pkspyder007, line 46: making out a helper utility make sense for re-using such logic

server/trpc/router/profile.ts
NiallJoeMaher: I'll refactor this later, I think there is a more elegant way to do this with the s3 client I'm just still learning it.

server/trpc/router/profile.ts
NiallJoeMaher: I'll pull it out if it becomes helpful in another part. I don't see it being used anywhere else for the moment 😁

pkspyder007 (approved):

PR #131, opened 2023-01-18
Feat/add basic cdk

NiallJoeMaher:
Adding CDK project      

- Adds s3, rds, cf, and domains
- Builds dockerfiles
- Adds test project to test pipeline first"

## Pull Request details:
Adds CDK project to get the infrastructure ready for alpha launch

## Any Breaking changes:
There could be a few when we start to switch to AWS.


cdk/bin/cdk-pipeline-codu.ts
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env node
+import "source-map-support/register";
+import * as cdk from "aws-cdk-lib";
+import { PipelineStack } from "../lib/pipeline-stack";
+
+const app = new cdk.App();
+
+new PipelineStack(app, "CoduPipeline", {
+  env: {
+    region: "eu-west-1",
```
pkspyder007: Move these to env variables

NiallJoeMaher: These aren't secrets weirdly but I wanted to obfuscate it so I just figured out how. 

<3

Issue #160, opened 2023-02-09
Add event for user signup on Fathom analytics


NiallJoeMaher:
## Context

Currently the only way to track new user signup is by checking the database. 

By adding an event to Fathom analytics on signup we should be able to track new user stats in the analytics 

## Expected Behavior

In Fathom we should be able to track new user signups

## Additional info

Docs [here](https://usefathom.com/docs/features/events).

EvanBurbidge:
PR opened

PR #170, opened 2023-02-17
Add fathom tracking to signup

EvanBurbidge:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
- Adds fathom tracking to the get-started sign up page.
- @NiallJoeMaher  can you tell me if this is the correct fathom id? or should it be a different one

https://github.com/codu-code/codu/issues/160

## Associated Screenshots:
    none

### DIFF
--- a/components/markdocNodes/Media/Media.tsx
+++ b/components/markdocNodes/Media/Media.tsx
@@ -83,9 +83,17 @@ export function Media(props: React.ReactPropTypes) {
   };
 
   const processCodePen = (mediaProps: MediaProps) => {
+    let src = mediaProps.src;
+
+    // Check if the URL contains /pen/ and replace it with /embed/
+    if (src.includes("/pen/")) {
+      src = src.replace("/pen/", "/embed/");
+    }
+
     type.current = MEDIA_TYPES.CODEPEN;
     return {
       ...mediaProps,
+      src
     };
   };
 


pages/get-started/index.tsx
```diff
@@ -43,17 +44,22 @@ const GetStarted: NextPage = () => {
...
-              await signIn("github", { callbackUrl: redirectTo });
+              try {
+                await signIn("github", { callbackUrl: redirectTo });
+                Fathom.trackGoal('sign-up-complete', 0);
+              } catch (e) {
+                console.error(e);
+              }
             }}
-            className="group relative w-full border border-transparent text-base bg-gradient-to-r from-orange-400 to-pink-600 rounded-md shadow-sm py-2 px-4 inline-flex justify-center font-medium text-white hover:from-orange-300 hover:to-pink-500 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-pink-300"
+            className="relative inline-flex justify-center w-full px-4 py-2 text-base font-medium text-white border border-transparent rounded-md shadow-sm group bg-gradient-to-r from-orange-400 to-pink-600 hover:from-orange-300 hover:to-pink-500 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-pink-300"
```
thewebuiguy: This is a built in linter I have for tailwind this stuff is done automatically

pages/get-started/index.tsx
```diff
@@ -43,17 +44,22 @@ const GetStarted: NextPage = () => {
           <button
             type="button"
             onClick={async () => {
-              await signIn("github", { callbackUrl: redirectTo });
+              try {
+                await signIn("github", { callbackUrl: redirectTo });
+                Fathom.trackGoal('sign-up-complete', 0);
```
NiallJoeMaher: We share a sign-in and sign-up page so this would trigger every time a user signs in. Can the client be added to the settings page when the user account is being setup in the `getServerSideProps`? I've not setup one of these trackers so I really don't know.

thewebuiguy: Will do!

package-lock.json
```diff
@@ -1,9601 +1,8 @@
 {
   "name": "codu",
   "version": "0.1.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
   "requires": true,
-  "packages": {
```
NiallJoeMaher: Any idea why this lock file is so different? 🤔

pages/settings/index.tsx
```diff
@@ -444,6 +444,7 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
   );
 
   if (!session?.user?.id) {
+    Fathom.trackGoal('sign-up-complete', 0);
```
NiallJoeMaher: Same issue as before since this doesn't confirm the user is new.

I think the only way to track this is during the initial account setup phase.

pages/get-started/index.tsx
NiallJoeMaher: What is that linter? Looks handy!

NiallJoeMaher (changes requested):

package-lock.json
thewebuiguy: Different npm versions?

pages/settings/index.tsx
thewebuiguy: Where is that in the code? the onSubmit method?

PR #192, opened 2023-03-06
fix(media tag): update media tag to format correct codepen url

xiaoniuniu89:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
- checks codepen url is correct format in media tag and updates if not

## Any Breaking changes:
- none


NiallJoeMaher (approved):
You are on fire! 🔥

PR #202, opened 2023-03-18
Added missing protocol in og:url meta tag for Articles

lmammino:
Missing the protocol might lead crawlers to think this is a relative URL. For example the FullStack bulletin crawler picked up an article URL as follows:

Screenshot 2023-03-18 at 10 59 17

This might also break previews for when articles are shared on social websites such as Twitter, Linkedin and Facebook.

For instance, Linkedin picks this up as HTTP:

Screenshot 2023-03-18 at 11 14 20

While Twitter doesn't seem to pick up the preview picture (not necessarily related but worth investigating too after this is fixed):

Screenshot 2023-03-18 at 11 15 02

### DIFF
--- a/pages/articles/[slug].tsx
+++ b/pages/articles/[slug].tsx
@@ -102,7 +102,7 @@ const ArticlePage: NextPage = ({
         <meta name="msapplication-config" content="browserconfig.xml" />
         <meta name="theme-color" content="#000" />
         <meta property="og:type" content="article" />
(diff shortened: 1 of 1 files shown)


NiallJoeMaher (commented):
Thanks for this! Points out the broken line but it's the `host` that is my issue. I'll fix this later. ❤️

NiallJoeMaher (approved):
Nope you were right! This is what I get for reviewing code while in the gym 😂 Thanks!

PR #280, opened 2023-05-11
fix: excerpt now generated from body if not provided

### LINKED ISSUE #259: Excerpts not being generated when left empty
## Expected Behavior

When a user creates a new post and publishes it without giving it an excerpt we should generate one from the start of the text.

## Current Behavior

This doesn't happen so there is a bug somewhere.

JohnAllenTech:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
Closes #259

Previously if the except text box was left blank the article would have no excerpt saved in the DB.

If the exceprt is not provided, This change will remove the markdown from the body and set the first 156 chars of the article as the default.

Note: I didnt populate the excerpt text box by default after clicking "Publish". This is to provide the user with the opportunity to write their own excerpt without having to delete the default one. Maybe this should be changed?

## Any breaking changes
None
## Associated Screenshots:
    
image
New article being written.

image
After clicking publish. Note no excerpt in text box

image
After clicking Publish again and navigating to Articles page. Note default excerpt generated

image
Second article being created

image
Excerpt populated

image
After clicking Publish again and navigating to Articles page. Note actual excerpt provided by user is used instead of default

### DIFF
--- a/pages/create/[[...postIdArr]].tsx
+++ b/pages/create/[[...postIdArr]].tsx
@@ -119,7 +119,7 @@ const Create: NextPage = () => {
       ...data,
       tags,
       canonicalUrl: data.canonicalUrl || undefined,
-      excerpt: data.excerpt || removeMarkdown(data.body.substring(0, 155), {}),
+      excerpt: data.excerpt || removeMarkdown(data.body, {}).substring(0, 155),
     };
     return formData;
   };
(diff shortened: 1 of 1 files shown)


NiallJoeMaher:
I'll review this a little later! Thanks for the help 🥇

NiallJoeMaher (approved):
This should work perfectly! Cheers

PR #282, opened 2023-05-14
feat: added new props to prompt service and updated style to match site

### LINKED ISSUE #203: Refactor promptService
## Context

the promptService is currently set up to only provide the service to one feature, the prompt that lets user know there are unsaved changes in their blog. It should be set up to be reusable to other components and take some props like <promptService message={}, prompt={} etc 

## Expected Behavior

a developer can use the promptService as a generic modal that takes in a few common arguments

## Current Behavior

the promptService is only set up to work with hard coded values for the blog editor to let user know they have unsaved changes.

JohnAllenTech:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
Closes #203 

This generalises the prompt factor a small bit.

Allows us to change the text displayed in the prompt through props.
This will allow this component to be reused in the future to support prevent moving away from other types of unsaved changes.

In addition to this. I took the opportunity while i was in the component to more closely align the styling with the rest of the white. This is entirely reversible if we do not want to go with the style changes.

Given I needed to change the bg colour for the modal component which was also being used for the delete article confirm check. I needed to also update the style there.

## Any Breaking changes:
Yes, if someone has a dev branch and is using the prompt service this will be a breaking change when they sync with develop.

I have fixed the breaking change within [[...postIdArr]].tsx for pages/create.

## Associated Screenshots:

   Prompt Service before changes
image
   Prompt Service after changes
image
   Delete article prompt before changes
image
   Delete article prompt after changes
image

***Note : They are the same size in all screenshots just different zoom levels I think***

### DIFF
--- a/components/Modal/Modal.tsx
+++ b/components/Modal/Modal.tsx
@@ -34,7 +34,7 @@ export function Modal({ open, onClose, children }: Props) {
               leaveFrom="opacity-100 translate-y-0 sm:scale-100"
               leaveTo="opacity-0 translate-y-4 sm:translate-y-0 sm:scale-95"
             >
-              <Dialog.Panel className="relative bg-white border-2 border-black px-4 pt-5 pb-4 text-left overflow-hidden shadow-xl transform transition-all sm:my-8 sm:max-w-lg sm:w-full sm:p-6">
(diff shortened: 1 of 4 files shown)


NiallJoeMaher (approved):
Looks great! We might make some variations of this one for light/dark mode too soon. I need to get the toggle switch in. Great job!

Issue #298, opened 2023-06-01
[Feature Request] Security Policy needed


CBID2:
## Context

The project's security  

## Expected Behavior

A full checklist

## Current Behavior

Hi @NiallJoeMaher! :) I just reviewed the project's [Community Standard Checklist](https://github.com/codu-code/codu/community) and noticed that it is not completely full due to missing a security policy
## Your Solution
I think we should add a security policy to the project because it will minimize the risk of receiving unauthorized commits and security is a top concern according to [this year's Open Source State Report](https://blog.opensource.org/the-2023-state-of-open-source-report-confirms-security-as-top-issue/). If approved, I'd be glad to implement this policy.

NiallJoeMaher:
Awesome idea! @CBID2, you have eagle eyes. Thanks for this. Would you mind implementing it still?

CBID2:
> Awesome idea! @CBID2, you have eagle eyes. Thanks for this. Would you mind implementing it still? 

Yep @NiallJoeMaher. I'll present it. Then, you give feedback and we'll take it from there.Sounds good?

NiallJoeMaher:
I didn't see this reply! Very happy to give feedback @CBID2. Thanks!!!

CBID2:
> I didn't see this reply! Very happy to give feedback @CBID2. Thanks!!!

Here it is @NiallJoeMaher: #306

PR #306, opened 2023-06-11
[feat] Create security policy

CBID2:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
- I created a security policy for this repository. It closes issue #298. 

## Any Breaking changes:

- None

## Associated Screenshots:
    
security policy

## Note to Reviewers
Hey @NiallJoeMaher! 😄 I didn't know what to do with the Supported Versions section, so I'm leaving that part for you to work on. 😄

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,17 @@
+# Security Policy
+
+## Supported Versions
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.1.x   | :white_check_mark: |
+| 5.0.x   | :x:                |
+| 4.0.x   | :white_check_mark: |
+| < 4.0   | :x:                |
+
+## Reporting a Vulnerability
+
+Please report (suspected) security vulnerabilities to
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,17 @@
+# Security Policy
+
+## Supported Versions
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.1.x   | :white_check_mark: |
```
NiallJoeMaher, line 7: What do these version things stand for in context of our app or is this just a template? 🤔

CBID2: Hi @NiallJoeMaher. It means the versions of the project that has been released. Think of it as Apple release a new version of their iOS products.

NiallJoeMaher (approved):
Since I need to do some reading on Security policies there is no point in holding this one up! We can update this when it makes sense. Thanks for this suggestion @CBID2 🦾

Issue #305, opened 2023-06-11
[Feature request] Show article count on users profile


NiallJoeMaher:
## Context

Currently there is no count for showing how many articles users have posted

## Expected Behavior

On the right-hand side of the header we should see the count of how many articles the user has posted. 
In the format of `(X)`.

## Current Behavior

No count - no fun

viniirbr:
@NiallJoeMaher could you assign this to me?

PR #309, opened 2023-06-14
feat: user's article count

serzhan181:
# ✨ Codu Pull Request 💻

## Pull Request details:
- Show a user's article on the right-hand side of the header. 
- Implementing https://github.com/codu-code/codu/issues/305

## Associated Screenshots:
изображение

### DIFF
--- a/components/Nav/Nav.tsx
+++ b/components/Nav/Nav.tsx
@@ -39,13 +39,13 @@ const Nav = () => {
     <Disclosure as="nav" className="bg-black">
       {({ open }) => (
         <>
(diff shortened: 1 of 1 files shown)


NiallJoeMaher:
Welcome @serzhan181! 🥳  I probably explained this one badly... I was hoping this would be only on the users profile like where I am showing in this screenshot:
Screenshot 2023-06-14 at 19 52 10

This also is live data - as in the real data and not a hard-coded value. So you could use the length of the posts to come up with a number.

serzhan181:
Oh ok i got it now!

PR #331, opened 2023-06-28
Improve images on sponsorship page, add padding to notification page title

### LINKED ISSUE #285: Update alpha/sponsorship page with webp images and use CSS to tilt them.
## Context
Currently we are using .png files for the images and they are chunky and slow. For optimising we should use .jpeg or webp image formats to stop the images being so large. 

## Expected Behavior

Images have a more suitable extension and faster load times/ 

## Current Behavior

Slow stuttering loading of images because they are too detailed for PNGs. 

## Screenshot
The page can be seen here: https://www.codu.co/alpha/sponsorship

### LINKED ISSUE #310: [Bugfix] Notifications list needs padding
## Context

Currently the body of the notification has no space to breathe from the title.
We should add a `mb-4` to the title area to give the notifications and/or no notification better spacing.

## Expected Behavior

Content is not squashed against the title section. 

## Current Behavior

See below

## Screenshots

Provide any additional information here
Screenshot 2023-06-14 at 19 57 43

sonngdev:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:

- Resolve #285:
  - Use CSS variable and `calc` function to rotate images instead of using inline style and React state.
  - Statically import images instead of loading from `public` directory to take advantage of Next's automatic image optimization.
  - Other: fix typo on sponsorship page: `mounth` -> `month`
- Resolve #310: See screenshots below

## Any Breaking changes:

None

## Associated Screenshots:
    
**Before**
Screenshot 2023-06-28 at 7 22 04 p m

**After**
Screenshot 2023-06-28 at 7 22 21 p m

### DIFF
--- a/pages/alpha/sponsorship/index.tsx
+++ b/pages/alpha/sponsorship/index.tsx
@@ -4,7 +4,7 @@ import Layout from "../../../components/Layout/Layout";
 import Image from "next/image";
 import Link from "next/link";
 import Head from "next/head";
-import { useEffect, useState } from "react";
+import { useEffect } from "react";
 
 interface Image {
   rotate: number;
@@ -41,16 +41,15 @@ const images: Image[] = [
 ];
 
 const Sponsorship: NextPage = () => {
-  const [scroll, setScroll] = useState(0);
-
   useEffect(() => {
     function handleScroll() {
-      setScroll(window.scrollY);
+      document.body.style.setProperty('--scroll', String(window.scrollY));
     }
     window.addEventListener("scroll", handleScroll);
 
     return () => {
       window.removeEventListener("scroll", handleScroll);
+      document.body.style.removeProperty('--scroll');
     };
   }, []);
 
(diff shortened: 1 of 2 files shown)


pages/alpha/sponsorship/index.tsx
```diff
@@ -69,27 +75,21 @@ const Sponsorship: NextPage = () => {
...
-                      ? `rotate(${(scroll / 200) * image.rotate}deg)`
-                      : `rotate(${image.rotate}deg)`
-                  }`,
+                  transform: `rotate(calc(min(var(--scroll), 200) / 200 * ${image.rotate}deg))`
                 }}
               >
                 <Image
                   src={image.src}
                   alt={image.alt}
-                  width={250}
```
NiallJoeMaher, line 91: Does this work without an explicit height/width? 🤔

sonngdev: @NiallJoeMaher It does. According to [the doc](https://nextjs.org/docs/app/building-your-application/optimizing/images#local-images), when using locally imported images...
> Next.js will [automatically determine](https://nextjs.org/docs/app/building-your-application/optimizing/images#image-sizing) the width and height of your image based on the imported file. These values are used to prevent [Cumulative Layout Shift](https://nextjs.org/learn/seo/web-performance/cls) while your image is loading.

However, image dimensions slightly change, because now their true dimension `268x268` is being used, instead of `250x250` as specified in the code.

**Before**
Screenshot 2023-06-28 at 10 23 51 p m

**After**
Screenshot 2023-06-28 at 10 22 33 p m

sonngdev: If we prefer the fixed `250x250`, I can add that back

sonngdev: I want to add that on client side, those images are converted to very lightweight `webp` (9x smaller than original images). This was already the case before, my change turn out to not bring additional optimization. 
Screenshot 2023-06-28 at 10 18 17 p m

So the chunky and slow feeling mentioned in #285 could solely be excessive re-rendering. Using CSS transformation should address this.

NiallJoeMaher (approved):
Looks good! 🎉

Issue #344, opened 2023-07-12
Show copy code block icon only on hover


praveen-quizizz:
## Expected Behavior

The copy icon should be displayed when the user hovers over the code block.

## Current Behavior

Currently the icon is always visible leading to poor reading experience.

## Screenshots

![image](https://github.com/codu-code/codu/assets/98312397/01a4dc86-8a13-4a6a-967f-85ff408c62b6)

## Steps to reproduce

Please provide detailed steps for reproducing the issue

1. Goto any article with code block e.g. https://www.codu.co/articles/setting-up-metacall-development-environment-on-windows-jxk6iltl

Shane-Donlon:
Hi Praveen,

Requirements update:

1. The copy / clipboard button should be visible for mobile.
2. The code excerpt / text should be completely visible to reduce side-scrolling.
3. On desktop the copy / clipboard button should be visible on mouse hover.


Thanks,
Shane

sudarshansb143:
@NiallJoeMaher Suggested enhancement here. Is the underlined empty space on code block is is needed or we can make `copy code` button positioning `absolute` so that it will remove this empty space ?
![image](https://github.com/codu-code/codu/assets/43947282/887b15de-eda4-4231-b76b-0603890c8c75)

Ex. 
![image](https://github.com/codu-code/codu/assets/43947282/a5a3ec81-2e4c-46aa-9121-8b9b8a296360)

NiallJoeMaher:
Thanks for that @sudarshansb143! We did that. 🥳

PR #352, opened 2023-07-22
#344 Show copy code block icon only on hover

Shane-Donlon:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
- Code block converted to grid to allow control of copy button (and code excerpt if needed)
- Text-wrap set for code text responsiveness.
- globals.css updated so the border goes around the code div (now grid parent) and not pre (code excerpt).
Note: I believe prettier updated quite a bit in the global.css see lines 84 / 85 for updates.
`.prose pre {
.prose .code {
  @apply border border-neutral-200 bg-neutral-100 dark:border-neutral-700 dark:bg-black;
}`
- Transform Scale Y used to "hide" the button as this is not resource intensive.
- Media query for copy button so that it  always visible on devices less than 1024px per the below docs. https://tailwindcss.com/docs/screens
- For devices bigger than 1024px hover and focus states added 
- A transition for the transform scale y has been added to grab users attention of the copy button (both hover and focus states).
- Tab index added to code & pre to allow users to skip to code sections / maneuver better using tab key.
- Button code has been moved up in the code div so that the DOM order is correct for screen-readers as this function is vitally important to the blog posts.
- General margin and padding updated for pre.

## Any Breaking changes:
None

## Associated Screenshots:

Text-Wrap
text wrap in code block

Copy button visible on mobile devices
code block mobile devices

Focus and hover states for desktop:
hover focus code block

### DIFF
--- a/components/markdocNodes/Code/Code.tsx
+++ b/components/markdocNodes/Code/Code.tsx
@@ -179,6 +179,10 @@ export default function Code({ children, language }: Props) {
             background: rgb(243 244 246);
             padding: 2px 2px 0px 4px;
           }
+
+          pre {
+            text-wrap: wrap;
+          }
         `}
       </style>
     </div>
--- a/components/markdocNodes/Code/Code.tsx
+++ b/components/markdocNodes/Code/Code.tsx
@@ -166,14 +166,13 @@ export default function Code({ children, language }: Props) {
       <style jsx>
         {`
           .code {
-            position: relative;
+            display: grid;
+            grid-template-rows: 30px 1fr;
           }
           .code button {
-            appearance: none;
-            position: absolute;
+            justify-self: end;
+            margin-right: 11px;
             color: inherit;
-            top: ${lines.length === 1 ? "17px" : "13px"};
-            right: 11px;
             border-radius: 4px;
             border: none;
             font-size: 15px;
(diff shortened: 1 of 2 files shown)


NiallJoeMaher:
@Shane-Donlon looks good! I'm just going to checkout the changes to make sure it works as expected. 🤝

Shane-Donlon:
@NiallJoeMaher 
Thanks Niall that would be great, I updated the branch yesterday and pushed the changes up but it's not finalized yet. Still need to correct light mode as I didn't test that. I'll put in a new PR once it's corrected, any feedback you have is great and I can implement into the new PR.

NiallJoeMaher:
> @NiallJoeMaher Thanks Niall that would be great, I updated the branch yesterday and pushed the changes up but it's not finalized yet. Still need to correct light mode as I didn't test that. I'll put in a new PR once it's corrected, any feedback you have is great and I can implement into the new PR.

That's perfect! I think everything looks good for the moment so I'll wait for your next PR. 

This article might help you quickly test the light/dark modes: https://www.codu.co/articles/easily-emulate-prefers-light-dark-mode-with-chrome-pszw3wei

NiallJoeMaher (approved):

PR #427, opened 2023-09-22
Fix bug #416 Random bookmarks

### LINKED ISSUE #416: Random bookmarks showing while not logged in
## Context

I haven't explored, but I have noticed random filled bookmarks on the homepage while logged out. 

Screenshot 2023-09-14 at 19 04 16

## Expected Behavior

There should be no bookmarks for people that are logged out since they don't have an account to bookmark things. 

## Current Behavior

Crazy stuff, go to the homepage logged out and you'll see there are bookmarks selected (they are blue when selected).

John-Paul-Larkin:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
What is happening is that any article that is liked by any logged in user, has the users id added to the Bookmarks array, which is a property of the Post object. When prisma queries the DB, it selects the userId from Bookmarks where userId. This seems to work fine when the user is logged in. However when userId is undefined it returns all the userId. Fixed by checking if user is undefined when assigning the currentUserLikesPost variable.

## Any Breaking changes:
None

## Associated Screenshots:
None    

Closes #416

### DIFF
--- a/server/trpc/router/post.ts
+++ b/server/trpc/router/post.ts
@@ -309,7 +309,8 @@ export const postRouter = router({
     });
 
     const cleaned = response.map((post) => {
-      const currentUserLikesPost = !!post.bookmarks.length;
+      let currentUserLikesPost = !!post.bookmarks.length;
+      if (userId === undefined) currentUserLikesPost = false;
       post.bookmarks = [];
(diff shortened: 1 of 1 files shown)


NiallJoeMaher (approved):
🌮 Thank you! Good work.

PR #460, opened 2023-10-02
feat: Added Light mode to Profile Page

### LINKED ISSUE #434: Profile page should have light/dark mode
## Context

So we can give a better user experience, we are working on a light/dark mode.

## Expected Behavior

We use Tailwind so we should enable this by having a default light mode and dark mode with the `dark:` prefix. 
You can see how this is done here: https://tailwindcss.com/docs/dark-mode

## Current Behavior

By default, we only use a dark theme. 
You can see the feed here: https://www.codu.co/settings

## Additional info

The background for light mode should be: `bg-neutral-100`
The text should be: `text-neutral-900`
The cards should have a background of: `bg-white`
borders for light mode should be: `neutral-100`

If there is more questions let me know as I'm sure I've missed something!

pateldivyesh1323:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
- Fixes : #434 
- Add Light mode to Profile page
## Any Breaking changes:
- None

## Associated Screenshots:

image
image

### DIFF
--- a/components/ArticlePreview/ArticlePreview.tsx
+++ b/components/ArticlePreview/ArticlePreview.tsx
@@ -146,7 +146,7 @@ const ArticlePreview: NextPage<Props> = ({
             {menuOptions && (
               <Menu as="div" className="relative">
                 <div>
-                  <Menu.Button className="p-1 rounded-full hover:bg-neutral-800">
+                  <Menu.Button className="p-1 rounded-full hover:bg-neutral-300 dark:hover:bg-neutral-800">
                     <span className="sr-only">Open user menu</span>
                     <DotsHorizontalIcon className="w-6 h-6" />
                   </Menu.Button>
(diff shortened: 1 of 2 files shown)


components/ArticlePreview/ArticlePreview.tsx
```diff
@@ -160,20 +160,20 @@ const ArticlePreview: NextPage<Props> = ({
                   leaveFrom="transform opacity-100 scale-100"
                   leaveTo="transform opacity-0 scale-95"
                 >
-                  <Menu.Items className="origin-top-right absolute bottom-10 right-0 mt-2 w-48 rounded-md shadow-lg py-1 bg-white ring-1 px-1 ring-black ring-opacity-5 focus:outline-none">
+                  <Menu.Items className="origin-top-right absolute bottom-10 right-0 mt-2 w-48 rounded-md shadow-lg py-1 bg-black ring-white dark:bg-white ring-1 px-1 dark:ring-black ring-opacity-5 focus:outline-none">
```
NiallJoeMaher: Could you reverse the styles and screenshot what it looks like? 
I didn't even think about the dark mode for the menu options. 🧠 

It looks good but I think it makes more sense to have a dark manu on the dark mode (and not the light mode). Thanks! 🙌

NiallJoeMaher (changes requested):

components/ArticlePreview/ArticlePreview.tsx
pateldivyesh1323: @NiallJoeMaher It looks like this.

image

NiallJoeMaher: This looks great! Let's get that pushed and I'll approve.

NiallJoeMaher (approved):
LGTM! ✨

PR #482, opened 2023-10-03
Authshifting

s0ham075:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
-  moved auth api route to app router 
- did some changes according to the previous review

## Any Breaking changes:
- Some pages might break since we havent moved all the pages to app router , alhough i have tried running the app and its working correctly
- One test is still failing for "should navigate to the sponsor page"

## Associated Screenshots:

- None

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "react-moveable": "^0.54.1",
     "react-textarea-autosize": "^8.5.3",
     "satori": "^0.10.4",
-    "sharp": "^0.32.5",
+    "sharp": "^0.32.6",
     "sonner": "^0.6.2",
     "styled-components": "^5.3.11",
     "superjson": "^1.13.1",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,8 @@
     "studio": "prisma studio",
     "migrate": "npx prisma migrate"
   },
-  "prisma": {    "seed": "npx tsx ./prisma/seed.ts"
+  "prisma": {
+    "seed": "npx tsx ./prisma/seed.ts"
   },
   "dependencies": {
     "@aws-sdk/client-s3": "^3.410.0",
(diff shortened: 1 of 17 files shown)


package.json
```diff
@@ -74,7 +74,7 @@
     "react-moveable": "^0.54.1",
     "react-textarea-autosize": "^8.5.3",
     "satori": "^0.10.4",
-    "sharp": "^0.32.6",
+    "sharp": "^0.32.5",
```
NiallJoeMaher: Is there a reason to downgrade this or just older code? 🤔

pages/index.tsx
```diff
@@ -45,7 +45,7 @@ const Home: NextPage = () => {
   return (
     <>
       <Head>
-        <title>Codú - The Web Developer Community</title>
+        <title>Codú -  Web Developer Community</title>
```
NiallJoeMaher: Could we revert this for the moment?

NiallJoeMaher (commented):
Really good work! thanks

pages/index.tsx
s0ham075: This was by mistake, i think i fixed this in the last commit.

package.json
s0ham075: This was by mistake too so I fixed that

s0ham075:
Thanks ! Should I make another pr or would you resolve the conflicts yourself?

NiallJoeMaher:
@s0ham075 if you could add a commit to this branch that would be great! There shouldn’t be much conflicts or that yet. 🦾

s0ham075:
Hey I think it should be fixed now

package.json
NiallJoeMaher: This still hasn't been installed back as the right version. If you need a hand let me know. 🦾

s0ham075: On it !

NiallJoeMaher (approved):
LGTM! Thanks 🦾 🌮

PR #483, opened 2023-10-04
Added custom hover and focus classes for social logos

DarrachBarneveld:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
Potential style change on footer social icons. Just felt a little creative in the evening.

## Any Breaking changes:
None

## Associated Screenshots:
    
footer

### DIFF
--- a/components/Footer/Footer.tsx
+++ b/components/Footer/Footer.tsx
@@ -74,11 +74,13 @@ const Footer = () => {
         <div className="mt-8 flex justify-center space-x-8">
           {navigation.social.map((item) => (
             <a
+              id={item.name.toLowerCase()}
               key={item.name}
               href={item.href}
               target="_blank"
               rel="noopener noreferrer"
-              className="text-neutral-600 hover:text-neutral-500 dark:text-neutral-500 dark:hover:text-neutral-400"
+              // className="text-neutral-600 hover:text-neutral-500 dark:text-neutral-500 dark:hover:text-neutral-400"
+              className="p-1 rounded-md"
             >
               <span className="sr-only">{item.name}</span>
               <item.icon className="h-6 w-6" aria-hidden="true" />
--- a/components/Footer/Footer.tsx
+++ b/components/Footer/Footer.tsx
@@ -74,13 +74,11 @@ const Footer = () => {
         <div className="mt-8 flex justify-center space-x-8">
           {navigation.social.map((item) => (
             <a
-              id={item.name.toLowerCase()}
               key={item.name}
               href={item.href}
               target="_blank"
               rel="noopener noreferrer"
-              // className="text-neutral-600 hover:text-neutral-500 dark:text-neutral-500 dark:hover:text-neutral-400"
-              className="p-1 rounded-md"
+              className={`p-1 rounded-md transition-all hover:scale-105 focus:scale-105 hover:brightness-110 focus:brightness-110 hover:text-white focus:text-white ${item.name.toLowerCase()}`}
             >
               <span className="sr-only">{item.name}</span>
               <item.icon className="h-6 w-6" aria-hidden="true" />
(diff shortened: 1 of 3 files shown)


components/Footer/Footer.tsx
```diff
@@ -74,11 +74,13 @@ const Footer = () => {
         <div className="mt-8 flex justify-center space-x-8">
           {navigation.social.map((item) => (
             <a
+              id={item.name.toLowerCase()}
               key={item.name}
               href={item.href}
               target="_blank"
               rel="noopener noreferrer"
-              className="text-neutral-600 hover:text-neutral-500 dark:text-neutral-500 dark:hover:text-neutral-400"
+              // className="text-neutral-600 hover:text-neutral-500 dark:text-neutral-500 dark:hover:text-neutral-400"
```
xiaoniuniu89: can remove comment please

styles/globals.css
```diff
@@ -365,6 +365,72 @@ blockquote {
...
+#discord:hover,
+#discord:focus {
+  transition: all 300ms;
+  transform: scale(1.05);
+  filter: brightness(1.1);
+  cursor: pointer;
+  color: white !important;
+  background: #4b83fb;
+  background: -moz-linear-gradient(top, #4b83fb, #734df8);
+  background: -ms-linear-gradient(top, #4b83fb, #734df8);
```
xiaoniuniu89: can we stick to one convention for colors, either hex or rgba. Better yet, use var names for maintainability - https://www.w3schools.com/css/css3_variables.asp

styles/globals.css
```diff
@@ -365,6 +365,72 @@ blockquote {
   color: #4b5563;
 }
 
+/* Footer social hover effects */
+
+#twitter:hover,
```
xiaoniuniu89: given the use of tailwind css can we move the styles from the global css file to inline tailwind styles?

components/Footer/Footer.tsx
xiaoniuniu89, line 1: looks good

styles/globals.css
DarrachBarneveld: This I am unsure how to do based on the alternating gradients. Well besides verbose if statements 🤔

xiaoniuniu89: https://tailwindcss.com/docs/hover-focus-and-other-states

https://tailwindcss.com/docs/reusing-styles

These might help. 

i am fine with using the css though. I would probably do the same. 

While using ids in global css will work, do you think it will be better for maintainability to move to a css module?

Something like Footer.module.css
// styles in here convert to classes ? since locally scoped we don't need an id now. class name will suffice. 

in Footer component import:
` import styles from './Footer.module.css';`

className={`p-1 rounded-md ${styles[item.name.toLowerCase()]}`}

something like this. I leave to you to choose implementation - your global styles will also work . 

https://nextjs.org/docs/app/building-your-application/styling/css-modules

NiallJoeMaher: Agreed with the use of tailwind. If you need a hand let me know and I'm happy to jump on a call and chat.

styles/globals.css
```diff
@@ -365,6 +365,72 @@ blockquote {
...
+#twitter:focus {
+  transition: all 300ms;
+  transform: scale(1.05);
+  filter: brightness(1.1);
+  cursor: pointer;
+  color: white !important;
```
NiallJoeMaher: We should be able to do some if not most of this with some Tailwind too.
Also, if possible remove the `!important`. The use of `!important` always causes bugs somewhere 😢

NiallJoeMaher (changes requested):
Really cool idea! And looks awesome. Thanks for opening up the PR. 🦾

styles/globals.css
DarrachBarneveld: I just made a pr, I used the @apply method, hope its an acceptable alternative.

NiallJoeMaher (approved):

Issue #498, opened 2023-10-08
Editor: Support for syntax highlighting in code block


xiaoniuniu89:
## Context

This is a story of master feature #197.

The editor is in components/editor.
## Expected Behavior

As a user, when I want to add a code snippet, there will be an option to select what programming language my snippet is in, so that my code example will look clearer.

## Current Behavior

There is no syntax highlighting feature.

## Steps to reproduce

Please provide detailed steps for reproducing the issue

1. open editor at /alpha/create/new
2. add a code block

## Additional info

https://tiptap.dev/examples/syntax-highlighting
https://tiptap.dev/api/nodes/code-block-lowlight

bhubneshmaharana:
hey i have implemented the required code . I am opening a PR .

Issue #497, opened 2023-10-08
Editor: Link tag missing _blank att in rendered html


xiaoniuniu89:
## Context

This is a story of master feature #197 
The editor is in components/editor

## Expected Behavior

Links should have an `_blank` attribute so they open in a new tab.

## Current Behavior

Links in the editor are fine. Links in rendered html (click preview in the editor) are missing this attr. Likely need to add to the config of Dom purifier in RenderPost.tsx

## Steps to reproduce

1. load the app and go to /alpha/create/new
2. add a link
3. click preview 
4. inspect the link in the dev tools, there will be no target="_blank" att.

## Additional info

https://www.npmjs.com/package/dompurify#can-i-configure-dompurify

DarrachBarneveld:
Would like to tackle a few of these editor issues 👍

NiallJoeMaher:
@DarrachBarneveld could these be split into a separate PR? 
Thanks 🦾

DarrachBarneveld:
No problem, will do soon

PR #507, opened 2023-10-09
Editor Changes

DarrachBarneveld:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
Added target blank and code highlight styling changes. Having issues rendering styled code to preview post. Install lowlight/highlight packages

## Any Breaking changes:
None

Assosiation with issue #498 and #497


components/editor/editor/CodeBlock.tsx
```diff
@@ -0,0 +1,52 @@
...
+        {extension.options.lowlight
+          .listLanguages()
+          .map((lang: string, index: number) => (
+            <option key={index} value={lang}>
+              {lang}
+            </option>
+          ))}
+      </select>
+
+      <pre className="bg-black border ">
```
xiaoniuniu89: @NiallJoeMaher are there styles for code blocks at the moment? may be easier to add here.

components/editor/editor/CodeBlock.tsx
```diff
@@ -0,0 +1,52 @@
+import { NodeViewContent, NodeViewWrapper } from "@tiptap/react";
+import { ChangeEvent, FunctionComponent } from "react";
+
+// Change this for code styling
+import "highlight.js/styles/monokai-sublime.css";
```
xiaoniuniu89: changes look good, check with @NiallJoeMaher re styling of the code blocks. Good job.

components/editor/editor/CodeBlock.tsx
xiaoniuniu89: Happy to deal with the smaller style issues at the end. Its not super important for alpha page

components/editor/editor/CodeBlock.tsx
DarrachBarneveld: I had quite a lot of difficulty getting the syntax highlighter to work in preview/render mode. Will take another look now

xiaoniuniu89: https://tiptap.dev/guide/node-views#render-html

you can extend the configuration for the node in tiptap extensions - . eg Codeblock.configure{renderHTML({ HTMLAttributes }) {
  return ['my-custom-node', mergeAttributes(HTMLAttributes)]
},}

Its not essential now, we can fix in another pr

DarrachBarneveld: I will try this, I went another way thats probably not ideal

DarrachBarneveld: The problem is its parses as a different node. The StarterKit.Codeblock one, not the customised one, 🤔

NiallJoeMaher: Is there any way to extend our current code blocks or is it just too different?

components/editor/editor/extensions/index.tsx
```diff
@@ -156,3 +185,8 @@ export const TiptapExtensions = [
     allowFullscreen: true,
   }),
 ];
+
+function highlightCode(code, language) {
+  const result = lowlight.highlight(language, code);
+  return `<pre><code class="hljs ${language}">${result}</code></pre>`;
```
xiaoniuniu89, line 191: Do we have to use this hljs class, can we copy our styles from our current code block style? 

I am happy to fix styles at a later date in any case.

@NiallJoeMaher

components/editor/editor/CodeBlock.tsx
xiaoniuniu89: you should be able to configure all of that. But If what you have works, then we can add it. I don't mind to fix styles up at the end.

components/editor/editor/CodeBlock.tsx
NiallJoeMaher: The styles we currently have are great; I wonder if we can mimic or what people think 🤔

components/editor/editor/CodeBlock.tsx
xiaoniuniu89: technically everything can be target in global.css with .ProseMirror code, .ProseMirror p etc to make the styles sync - we can figure it out later if you wish, one source of truth should be feasable in the editors case

DarrachBarneveld: I have two solutions. Maybe I should PR the alternate also, each have pros and cons I think?

components/editor/editor/CodeBlock.tsx
xiaoniuniu89: We should be able to - We can add this and figure it out - if it doesnt work we can revert the changes and save syntax highlighting for v2 of the editor

components/editor/editor/CodeBlock.tsx
NiallJoeMaher: It might be worth chatting about it in the Discord too just so we can discuss too!

DarrachBarneveld: Ok, will log on this evening 👍

NiallJoeMaher:
Let me know if you are going to fix the conflicts and we can add it in case we need backward compatibility.

NiallJoeMaher:
If you can would you fix the conflicts and I'll get this merged! I also need to send you the Codú Hacktoberfest badge 🦡

DarrachBarneveld:
OMG! A collectible badge, my weakness. Will fix conflicts now 😃

NiallJoeMaher (approved):
LGTM! 🌮

PR #513, opened 2023-10-11
add all formatting options to slash menu

John-Paul-Larkin:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
Added commands to slash menu. 
I did not add the table command as it is a work in progress.

The main toolbar has both a toggleCode and toggleCodeBlock command, which is a bit confusing. Maybe there is a distinction between the two?
As the slash menu already has the toggleCodeBlock, I left that in.

## Any Breaking changes:
None

### DIFF
--- a/components/editor/editor/extensions/slash-command.tsx
+++ b/components/editor/editor/extensions/slash-command.tsx
@@ -21,6 +21,17 @@ import {
   TextQuote,
   Image as ImageIcon,
   Code,
+  YoutubeIcon,
+  BoldIcon,
+  RectangleHorizontalIcon,
+  ItalicIcon,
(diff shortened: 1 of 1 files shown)


NiallJoeMaher:
This is going to be very frustrating, but I was discussing with @xiaoniuniu89 about stripping out nearly all the features 🙈 

The only features we will have in the first version is the same as Medium to give a simple experience (and a consistent look for articles). 

Thanks for the initiative on this one but let me see if I can come up with some more issues in this line for a total cleanup.

John-Paul-Larkin:
No worries. All part of the process! 
Ill close this pr so.

PR #525, opened 2023-10-13
Alteditor - 

DarrachBarneveld:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:

This is the alternate solution to rendering coding highlighted syntax.

Differences/Pros/Cons

1. No use of read only editor instance to create html layout
2. Dangerously inserts html
3. More concise code
4. Manually manipulates DOM instead of allowing React to do this

The pros are its probably easier to render old markdown posts. Its less code and only requires a single component. The con is the preview will never be an exact match of the editor view as with a read only editor.

## Any Breaking changes:
None

Let me know which option is preferable. I kind of like this one slightly more but a readonly editor instance will mean we will never have to worry about styling issues across all the posts.


NiallJoeMaher:
Closing in favour of #507

PR #545, opened 2023-10-14
Feature 168 report posts

John-Paul-Larkin:
# ✨ Codu Pull Request 💻

## Pull Request details:
This is the updated report comment feature . Should not require much review, as it basically all the same code form the previous PR. The only difference is the trpc router parts.

Need to add an ADMIN_EMAIL to the .env file.

### DIFF
--- a/components/Comments/ReportComments.tsx
+++ b/components/Comments/ReportComments.tsx
@@ -1,0 +1,165 @@
+import React, { useRef, useState, useEffect } from "react";
+import Flag from "../../icons/flag.svg";
+import { XIcon } from "@heroicons/react/outline";
+import toast from "react-hot-toast";
+import { signIn, useSession } from "next-auth/react";
+import { createReportEmailTemplate } from "../../utils/createReportEmailTemplate";
+import { trpc } from "@/utils/trpc";
+// import { trpc } from "../../utils/trpc";
+import { api } from "@/server/trpc/react";
+
+interface Props {
+  name: string;
+  body: string;
+  id: number;
+  email: string | null;
+  slug: string;
+}
+
+export const ReportComments = (props: Props) => {
+  const { mutate: sendEmail } = api.emailReport.send.useMutation({
+    onSuccess: () => {
+      toast.success("Report sent");
+    },
+    onError: () => {
+      toast.error(
+        "Oops, something went wrong, please send us a message on github https://github.com/codu-code/codu",
+      );
+    },
+  });
+
+  const { data: session } = useSession();
+  const { name, body, id, email, slug } = props;
+  const [isModalOpen, setModalOpen] = useState<boolean>(false);
+  const [comment, setComment] = useState("");
+
+  const handleOpenModal = () => setModalOpen(true);
+  const handleCloseModal = () => setModalOpen(false);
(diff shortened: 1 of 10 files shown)


NiallJoeMaher:
I'll need to test this code a little since it'll result in emails to me 🙈 I'll try get to it later today! Thanks. 🦾

John-Paul-Larkin:
No problem. There is a good few changes.

server/api/router/emailReport.ts
```diff
@@ -0,0 +1,23 @@
+import toast from "react-hot-toast";
+import { SendEmailSchema } from "../../../schema/comment";
+import sendEmail from "@/utils/sendEmail";
+
+import { createTRPCRouter, protectedProcedure } from "../trpc";
+
+export const emailReportRouter = createTRPCRouter({
+  send: protectedProcedure
```
NiallJoeMaher, line 8: Do we tell a user to login if they want to report a post?

NiallJoeMaher: And if not maybe this will be an enhancement later. 🦾

John-Paul-Larkin: onClick={() => (session ? handleOpenModal() : signIn())}

I handled this on the button. If there is no session the auth signIn() funciton is called. 
Is that the right way to handle it?

components/Comments/ReportComments.tsx
```diff
@@ -0,0 +1,163 @@
...
+    };
+    document.addEventListener("mousedown", checkIfClickedOutside);
+    return () => {
+      document.removeEventListener("mousedown", checkIfClickedOutside);
+    };
+  }, [isModalOpen]);
+
+  return (
+    <>
+      <button
```
John-Paul-Larkin, line 99: Just figured out how to comment on a particular line :)

NiallJoeMaher (approved):
Nice one! I think I'll open up an issue to create one for blog posts too in case you want to take it 😉 Thanks a million! 🌮

John-Paul-Larkin:
> Nice one! I think I'll open up an issue to create one for blog posts too in case you want to take it 😉 Thanks a million! 🌮

Blogs as in Articles? 
I had thought about it, but I wasnt sure. I could easily add the same flag button at the bottom of the article, just modify the props for the email. Or I could make it into an actual button, with label 'report blog'. Just to differentiate it.

BTW if you test the email and all goes well could you forward it to me, johnplarkin@gmail.com thanks

NiallJoeMaher:
> > Nice one! I think I'll open up an issue to create one for blog posts too in case you want to take it 😉 Thanks a million! 🌮
> 
> 
> 
> Blogs as in Articles? 
> 
> I had thought about it, but I wasnt sure. I could easily add the same flag button at the bottom of the article, just modify the props for the email. Or I could make it into an actual button, with label 'report blog'. Just to differentiate it.
> 
> 
> 
> BTW if you test the email and all goes well could you forward it to me, johnplarkin@gmail.com thanks

I'll tag you in an issue later. I think we can use the 3 dots menu on the options bar and add "Report post" as one of the options from the menu. Simple and takes up no more space. In fact that might be nice pattern for anywhere we need it. 😅

PR #548, opened 2023-10-14
Debounce the search bar search

### LINKED ISSUE #487: Debounce the search bar search
## Context

Currently, the `/articles` page has a search bar (in `develop`, not live yet). 
Each keystroke currently starts a search. 
We should debounce the value so that the search only happens after a user stops typing. 

## Expected Behavior

I expect the value to be debounced for 500ms to limit the amount of requests. 

## Current Behavior

A search is requested on every keystroke.

MadhavKrishanGoswami:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
  Debounce the value  so that the search only happens after a user stops typing and that will limit the amount of requests.
## Any Breaking changes:
   None

## Associated Screenshots:
    None

Fixes #487

### DIFF
--- a/components/ArticleSearch/SearchBar.tsx
+++ b/components/ArticleSearch/SearchBar.tsx
@@ -1,18 +1,35 @@
+import React, { useState, useEffect } from "react";
+
 interface Props {
   searchTerm: string;
   setSearchTerm: React.Dispatch<React.SetStateAction<string>>;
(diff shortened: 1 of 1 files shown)


NiallJoeMaher (approved):
LGTM! Good job. 🦾

PR #584, opened 2023-10-21
Fix styles

CarolinaCobo:
# ✨ Codu Pull Request 💻

Codu Logo
## Pull Request details:
- Change ring colours from rose to pink-500 and from pink-300 to pink-600
- Added new default styles in the Global styles file
- Focus visible (when navigating with keyboard) to match the styles of the rest of the page. 

## Any Breaking changes:
None

## Associated Screenshots:
Screenshot 2023-10-21 at 13 44 23

Screenshot 2023-10-21 at 13 45 04

Screenshot 2023-10-21 at 13 45 44

Screenshot 2023-10-21 at 13 46 05

### DIFF
--- a/styles/globals.css
+++ b/styles/globals.css
@@ -37,25 +37,37 @@ body {
     @apply mt-1 bg-white dark:bg-black focus:outline-none focus:ring-2 focus:ring-neutral-300 ring-offset-1 block w-full shadow-sm sm:text-sm border-white border-2 px-2 py-2 disabled:opacity-50 text-black dark:text-white;
   }
 
+  a {
+    @apply focus-style p-1;
+  }
+
   label {
     @apply block text-sm font-medium text-white;
   }
 }
 
 .nav-button {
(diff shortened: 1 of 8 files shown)


styles/globals.css
pkspyder007, line 1: @CarolinaCobo I think moving the styles to a button component will be better since these classes are not where documented we'll loose track of them.

Having a `Button` component that can render multiple variants would be a good approach.

pkspyder007 (commented):
Left one comment for discussion over global styles vs component level variants

styles/globals.css
NiallJoeMaher: Good point @pkspyder007! I think we will open another issue/discussion on this though and it might need a little refactor. 🦾

CarolinaCobo: @pkspyder007 thank you for your comment! I moved the main button into its own existing class and removed the white line around. I'll chat to @NiallJoeMaher about doing what you suggested, as it seems like a better approach.

NiallJoeMaher (approved):
LGTM 🌮

PR #587, opened 2023-10-22
Use session correctly and styles in the side nav

CarolinaCobo:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
- Destructure the data for the session so the saved posts only show if the user is authenticated.
- In light mode show the correct color on the side navigation. 

## Any Breaking changes:
- None

## Associated Screenshots:
 
Screenshot 2023-10-22 at 20 14 22


NiallJoeMaher (approved):
🌮

PR #589, opened 2023-10-22
fix-scrollbar-on-comments 

### LINKED ISSUE #543: Scroll bar on comments
## Context
On chrome there is a scroll bar below every comment on the article page. This does not look good, especially when there are a numerous comments.

## Expected Behavior
Remove/hide scrollbar.

## Screenshots
In this image chrome is on the left, firefox is on the right. I have not tested any other browsers

Screenshot from 2023-10-14 11-02-03

vrun1208:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
- fixed scroll in the comments section which shows horizontal scroll-bar on every comment. As we were using property `overflow-x-scroll` which works well in chrome. But there is no need of scroll in the comments as the div's are working as intended. So, I replaced the `overflow-x-scroll` with `overflow-x-hidden`.
## Any Breaking changes:
None

## Associated Screenshots:
image
fix: #543

### DIFF
--- a/components/Comments/CommentsArea.tsx
+++ b/components/Comments/CommentsArea.tsx
@@ -294,7 +294,7 @@ const CommentsArea = ({ postId, postOwnerId }: Props) => {
                 </div>
 
                 <div className="border-l-2 border-neutral-700 ml-4 pl-2 -mt-2">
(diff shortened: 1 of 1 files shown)


NiallJoeMaher (approved):
🌮  LGTM!

PR #590, opened 2023-10-22
Fix styles in buttons when hovering and in dark and light

CarolinaCobo:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
- Matching colour in both buttons both in light/dark mode.
- Add the discloser wrapper when logging in.
- Remove group focus so the Icon doesn't stay hoovered. 

## Any Breaking changes:
- None

## Associated Screenshots: 

Screenshot 2023-10-22 at 23 36 14


components/Nav/Nav.tsx
```diff
@@ -139,11 +139,11 @@ const Nav = ({ session }: { session: Session | null }) => {
                       <Link
```
pkspyder007, line 137: @CarolinaCobo looks like there is no gap between the bell icon and theme toggle, let add that too. Everything else looks good.

NiallJoeMaher (changes requested):
I think we need to pass the Link element to this so we get the benefits. I can add a commit if you don't have time

NiallJoeMaher (approved):
LGTM! 🌮

PR #599, opened 2023-10-24
Fix: Report backdrop now hides all elements

### LINKED ISSUE #595: When opening report modal the menu shouldn't be above the modal background
## Context

When opening the report modal the side menu and options are very visible and should be hidden behind the modal backdrop like the rest of the page,

Here's a screenshot that shows the current behaviour:

Screenshot 2023-10-24 at 11 31 53

Make sure this behavior is the same on mobile too! 🦾 

## Expected Behavior

The menu should be behind the modal background and not on top like the Modal. 

## Steps to reproduce

1. Go to any article
2. Select the menu on the left
3. Select "Report article" to view the modal

LemonDrop847:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:🤜
- Fixes: #595 
- Moved the report modal up in relative z.

## Any Breaking changes:
- Seems to be none.

## Associated Screenshots:
- Before:
z
- After:

- Desktop:
x
- Mobile:
y

### DIFF
--- a/components/ArticleMenu/ArticleMenu.tsx
+++ b/components/ArticleMenu/ArticleMenu.tsx
@@ -109,7 +109,7 @@ const ArticleMenu = ({
       enterFrom="transform opacity-0 scale-75"
       enterTo="transform opacity-100 scale-100"
     >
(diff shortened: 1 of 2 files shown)


NiallJoeMaher (approved):
Very quick work! Thanks 🌮

PR #643, opened 2023-11-20
[Codú - #642] - Update theme provider

### LINKED ISSUE #642: Dark/Light Mode not all reactive to system preferences
## Context

The current implementation of the theme provider says to enable system color preferences but these values don't get provided because the default value is set to "dark".

## Expected Behavior

If you enable system prop as true, which it is by default according to [next-themes](https://www.npmjs.com/package/next-themes), then you can utilise the system preferences for the theme. But when the user changes the theme manually use the local storage value provided by the next-themes useTheme hook.

1. Load UI and use system preference
Screenshot 2023-11-20 at 18 09 13

2.Update system preference and observe no update to local storage
Screenshot 2023-11-20 at 18 09 53

3. Update manually and see local storage update to persist values for the future
Screenshot 2023-11-20 at 18 10 49
Screenshot 2023-11-20 at 18 10 52

## Current Behavior

The default value gets set to dark, instead of "system" in local storage and is used as the default thereafter.

1. Observe dark theme by default
codu with dark mode as default

3. System update not applied
change system preferences to light

5. Only manual updates work
light more preference selected

garyb1:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:

Closes #642 

If you enable system prop as true, which it is by default according to [next-themes](https://www.npmjs.com/package/next-themes), then you can utilise the system preferences for the theme. But when the user changes the theme manually use the local storage value provided by the next-themes useTheme hook.

## Any Breaking changes:
None

## Associated Screenshots:
    
Screenshot 2023-11-20 at 18 09 53
Screenshot 2023-11-20 at 18 10 49
Screenshot 2023-11-20 at 18 10 52

### DIFF
--- a/components/Theme/ThemeProvider.tsx
+++ b/components/Theme/ThemeProvider.tsx
@@ -1,14 +1,10 @@
 "use client";
 
-import type { ReactNode } from "react";
+import { type ReactNode } from "react";
 import { ThemeProvider } from "next-themes";
 
 const Providers = ({ children }: { children: ReactNode }) => {
-  return (
-    <ThemeProvider attribute="class" defaultTheme="dark" enableSystem>
-      {children}
-    </ThemeProvider>
-  );
+  return <ThemeProvider attribute="class">{children}</ThemeProvider>;
 };
 
 export default Providers;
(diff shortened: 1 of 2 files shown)


NiallJoeMaher (approved):
Looks great! Thanks for this. 🔥

Issue #505, opened 2023-10-08
Editor: fixed menu should hide when clicking publish


xiaoniuniu89:
## Context

We have a fixed menu at the top of our new editor

## Expected Behavior

when user clicks the publish button, the menu should not be visable

## Current Behavior

visable from publish screen

## Screenshots

<img width="1271" alt="Screenshot 2023-10-08 at 21 46 39" src="https://github.com/codu-code/codu/assets/84312595/4f8960c8-675d-4606-9703-937d904c22aa">

## Steps to reproduce

Please provide detailed steps for reproducing the issue

go to alpha/create/new
click publish button

John-Paul-Larkin:
Hey Niu. I will try this tomorrow.

NiallJoeMaher:
@Fishamble are you taking this one?

VickTor61:
Hi @NiallJoeMaher . I will like to contribute to this or fix any other ticket that hasn't been resolved yet

John-Paul-Larkin:
Hi Vicktor. I am not working on this, so you can give it a try.

vrun1208:
Hi @NiallJoeMaher, looks like the issue is with css, its been a week there is no activity in this issue. If no one is working I can open a PR!

NiallJoeMaher:
@VickTor61 are you working on this or can I pass it over to @vrun1208?

VickTor61:
@NiallJoeMaher  currently working on it I should make a PR later today

VickTor61:
@NiallJoeMaher The error doesn't show in development, Has it already been fixed?
![Screenshot 2023-10-25 at 5 43 03 AM](https://github.com/codu-code/codu/assets/63825684/ac60f441-9157-4eea-b72e-0e489e285090)

John-Paul-Larkin:
Hi Vick. You are looking at the wrong editor. 
http://localhost:3000/alpha/new is the url for the alpha editor.
app/alpha/new/[..postidarr]/client.tsx is the component.

danboyce92:
Hey there, I have a solution for this problem put together. Could I have this assigned to me so I can create a PR?

vrun1208:
I already fixed this issue earlier! @NiallJoeMaher can I open a PR!

PR #682, opened 2024-01-02
Toolbar is hidden while publishing an article

danboyce92:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details:
*This PR is in response to issue #505*

In this PR I added a piece of state 'toolbarVis' in the useCreatePage hook. From there I had it imported into the alpha/new client.tsx where it was passed down as props into the editor component and from there it was passed into the Toolbar component. It is used in the Toolbar component to swap the css class whenever the 'open' state is changed.

There is an additional class added to the bottom of the Toolbar.module.css called .hide
which changes the visibility to hidden;

The 'open' state is monitored in the alpha/new client.tsx with an additional useEffect which changes the state of toolbarVis depending on if the publish button has been clicked or not.

## Any Breaking changes:
**None** as far as I'm aware everything seems to be working fine.
I'm aware it's probably not ideal passing state through multiple components before it gets to where it's needed as it can get messy fast. Usually I would use a redux store. If there is a simpler solution please let me know and I'll get to work.

 

## Associated Screenshots:
    

pr1
pr2

### DIFF
--- a/app/alpha/new/[[...postIdArr]]/_client.tsx
+++ b/app/alpha/new/[[...postIdArr]]/_client.tsx
@@ -21,6 +21,8 @@ const Create = () => {
     savedTime,
     open,
     setOpen,
+    toolbarVis,
+    setToolbarVis,
     hasUnsavedChanges,
     handleSubmit,
     register,
@@ -46,6 +48,14 @@ const Create = () => {
   useEffect(() => {
     _setUnsaved(hasUnsavedChanges);
   }, [hasUnsavedChanges, _setUnsaved]);
+
+  useEffect(() => {
+    if (open) {
+      setToolbarVis("styles.hide");
+    } else {
+      setToolbarVis("styles.sticky");
+    }
+  }, [open])
 
   return (
     <>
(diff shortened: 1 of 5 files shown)


NiallJoeMaher (changes requested):
This works great but there could be an easier solution if you don't mind the change request? If you don't have time let me know and I'll merge it and open up an issue with this comment.

If you make the toolbar have a lower `z-index` we might avoid all the logic. I'm not sure how we do that off the top of my head. Here are the results by just toggling z-index:

Screenshot 2024-01-02 at 21 59 44
Screenshot 2024-01-02 at 21 59 51



**If this proves too difficult here's a nuclear option:**
I'm also open to removing the `fixed` property as this toolbar is temporary until we have the new toolbar which will open when you highlight text like this:
Showing a toolbar on hover
I'm playing with designs here https://www.figma.com/file/yKc7ind3lHUTpqkjTePy1N/codu.co?type=design&node-id=0%3A1&mode=design&t=ZrC8BZjxWoPsnPAn-1

danboyce92:
Yeah that z-index approach is way simpler, with less mess! Made the change there in the second commit. Good stuff!

NiallJoeMaher (approved):
LGTM! 🌮

Issue #702, opened 2024-01-25
Feature: Schedule articles


NiallJoeMaher:
It would be great to be able to schedule articles so that people can stack posts for consisting posting.

PR #704, opened 2024-01-30
Schedule post

### LINKED ISSUE #689: Feature: Schedule articles
## Context

I think it would be useful if we could schedule articles so that I can batch create articles and schedule them for regular intervals

## Expected Behavior

User can schedule articles to be released in the future.
I will look for patterns on doing this but my initial feeling is that if we add a published date in the future we should be able to just hide all posts until the published time has hit since currently we use a timestamp at publish time. 

## Current Behavior

The options are to save and publish. There is no schedule option.

sonngdev:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details

✅  Close #689 and #702 
    - Implement post scheduling feature using the field `published` of the table `Post` (see discussion in #689).
    - `/articles` page now only includes posts that are considered published according to the new logic.
    - `/create` page now supports post scheduling (see screenshots).
    - `/my-posts` page now includes an additional tab `Scheduled`. Additionally, published time and scheduled time are also added in respective tabs (see screenshots).
🐞  Fix some existing bugs on `/create` page.
    - Fix incorrect id for tag input.
    - Fix redirection after submitting form (use `router.push` because `redirect` doesn't work with client component).
    - Fix toast when post fails to submit.
    - Improve some button label text.

## Any Breaking changes

New feature, some stuff was renamed. Shouldn't break anything, but some alpha components should be updated later to include this new feature (`/app/alpha/new/[[...postIdArr]]/_client.tsx` for example).

## Associated Screenshots

image
<hr>
image
<hr>
image
Using browser's native date time picker. A scheduled post can be re-scheduled as long as it has not hit published time yet.
<hr>
image
Scheduled posts are sorted by ascending schedule (published) time.
<hr>
image
Published time is added to the bottom of the card. Sorting order doesn't change.

### DIFF
--- a/components/Switch/Switch.tsx
+++ b/components/Switch/Switch.tsx
@@ -1,0 +1,22 @@
+"use client";
+
+import * as React from "react";
+import * as SwitchPrimitives from "@radix-ui/react-switch";
+
+const Switch = React.forwardRef<
+  React.ElementRef<typeof SwitchPrimitives.Root>,
+  React.ComponentPropsWithoutRef<typeof SwitchPrimitives.Root>
+>(({ className, ...props }, ref) => (
+  <SwitchPrimitives.Root
+    className={`peer inline-flex h-6 w-11 shrink-0 cursor-pointer items-center rounded-full border-transparent from-orange-400 to-pink-600 transition-colors disabled:cursor-not-allowed disabled:opacity-50 data-[state=unchecked]:bg-neutral-300 data-[state=checked]:bg-gradient-to-r dark:data-[state=unchecked]:bg-neutral-700
+      ${className}
+    `}
+    {...props}
+    ref={ref}
+  >
+    <SwitchPrimitives.Thumb className="pointer-events-none ml-[2px] block h-5 w-5 rounded-full bg-black shadow-lg ring-0 transition-transform data-[state=checked]:translate-x-5 data-[state=unchecked]:translate-x-0 dark:bg-white" />
+  </SwitchPrimitives.Root>
+));
+Switch.displayName = SwitchPrimitives.Root.displayName;
+
+export { Switch };
--- a/components/Switch/Switch.tsx
+++ b/components/Switch/Switch.tsx
@@ -14,7 +14,7 @@ const Switch = React.forwardRef<
     {...props}
     ref={ref}
   >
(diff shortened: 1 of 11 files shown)


utils/post.ts
sonngdev, line 1: @NiallJoeMaher I am not sure if creating a new file is appropriate for this helper utility. Let me know if this needs changing.

components/Switch/Switch.tsx
sonngdev, line 1: Creating a new UI component. I think we may need a design system soon.

NiallJoeMaher: Do you think it would be better without a toggle? Maybe like if it's populated or not? With a clear button?

Another question, does this make sure we can't post something in the past?

I'm reviewing on my phone so I could be missing it. 😅

sonngdev: @NiallJoeMaher 

You are right, I forgot to handle that case. Will push some more code soon.

We could do it without a toggle for sure. I'm just coming up with a proposal for the UI here, and actually the toggle was inspired by TikTok. 

In my opinion, having a toggle could be clearer than having an empty/populated input, but since you will likely use this feature much more than me, I'm open to your input.

sonngdev: Added a commit to handle past schedule time

NiallJoeMaher (changes requested):
One last thing I think 🤔 But I think the `[username]` will need to be updated too!

sonngdev:
> One last thing I think 🤔 But I think the `[username]` will need to be updated too!

@NiallJoeMaher Can you be more specific? I don't get what you are referring to, because AFAIK this PR doesn't change anything related to username.

Also, I know this is a large PR, so no pressure to merge quickly, boss.

NiallJoeMaher:
> > One last thing I think 🤔 But I think the `[username]` will need to be updated too!
> 
> @NiallJoeMaher Can you be more specific? I don't get what you are referring to, because AFAIK this PR doesn't change anything related to username.
> 
> Also, I know this is a large PR, so no pressure to merge quickly, boss.

For example, if I got to https://www.codu.co/niall I see all of my articles. Every user has all of their own posts on their page.

NiallJoeMaher:
@thanhsonng let me know if the last comment makes sense. 🦾 Thanks!

sonngdev:
> @thanhsonng let me know if the last comment makes sense. 🦾 Thanks! 

@NiallJoeMaher yes it does, thanks for the example it really cleared things up. I will add a couple more commits soon.

sonngdev:
@NiallJoeMaher Updated, thanks for catching it, I totally missed that [username] page. That's why code reviews are awesome 😎 If I miss anything else just let me know.

NiallJoeMaher (approved):
LGTM! Nice one 🦾

PR #710, opened 2024-02-04
Implement deleteDataFromAllTables function

John-Paul-Larkin:
# ✨ Codu Pull Request 💻

Codu Logo

Added a function which deletes all data, from all tables in the database.
This is invoked before the database is filled with the seed data.

I tried adding a user prompt before the deletion process. 
I was unable to implement synchronous (or async) command line entry.
I am not sure why exactly, but my assumption is that it is something to do with the way the prisma CLI handles asynchronous operations in the 'npx prsima db seed' script.
 
## Any Breaking changes

None

## Associated Screenshots

None


NiallJoeMaher (approved):
Nice! Looks like this will work well 🦾

PR #713, opened 2024-02-04
redirect user to login page

John-Paul-Larkin:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details

Resolve issue which redirected users to an error page, instead of the sign-in page, upon clicking the bookmark button.


NiallJoeMaher (changes requested):
Pedantic, but could you update this to be:

```
if (!session) {
  return signIn();
};
if (bookmarked) {
  return bookmarkPost(id, false);
}
bookmarkPost(id);
```
Having the logical closures is a little easier to read.

NiallJoeMaher (approved):
Nice and quick! Thanks! 🌮

PR #727, opened 2024-02-14
Feature onboard developer form

John-Paul-Larkin:
# ✨ Codu Pull Request 💻

## Pull Request details

- Adds a  form for developers to enter their details if they are interested in job opportunities. 

-  Added a new prisma model for the developer details, so migration required

- npx prisma migrate dev

I need to consult with you about a few issues before merging.
- Styling
- Further data to be added location / CV / Do you wish to be contacted option etc.
- Few other bits, maybe we can do a call? 

## Any Breaking changes

Possibly, as I had to remove the id @unique constraints  in the membership and RSVP models.

## Associated Screenshots
Sorry about the sound quality
https://youtu.be/5Foq5QHGh4o


NiallJoeMaher (changes requested):
Let's have a call this morning when/if you are free!
As always, the video is so helpful. Saves my ass. 🦾 
Love the logic checks in Zod! Looks like it was a pain. 😂 

**This is a list to remind me:**
- I think we should be able to find a package for the countries most likely.
- The progress bar should be the entire screen width
- I think we should just move all the data to under `user` instead of `developerDetails`. 🦾  This might cure a lot of the nesting issues too.
- The `@unique` error, maybe we need to let the database generate the IDs but I'll look into it. 
- Add gap between input boxes (flex gap) for date of birth
- We can probably drop the review data in the final step when we migrate this to the production build but really helpful for while we test it. 
- Instead of using state for which slide we are on we could use a query param (just so we  can link to particular pages). And also, as a fallback if it doesn't exist go to slide `1`.
- I think we need to validate the data inside the action too so that we can validate on the serverside too. You should be able to share the schemas from Zod and do the same check really.

John-Paul-Larkin:
Nice one, thanks for the feedback.
Ill go through these and do another PR.
The one that jumps out to me is :
"Instead of using state for which slide we are on we could use a query param (just so we can link to particular pages). And also, as a fallback if it doesn't exist go to slide 1"
I think I can make the whole thing a SSR page in that case.

Also :
"I think we need to validate the data inside the action too so that we can validate on the serverside too."
Cant believe I missed this :upside_down_face:

John-Paul-Larkin:
I implemented most things, but I left in the review data screen for now.

https://youtu.be/IC-s_G0fD84

PR #734, opened 2024-02-17
fix: added validation for username (only alphanumeric and dashes allowed)

Maczi01:
…wed)

# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details

Added validation in saveSettingsSchema (schema/profile.ts). I used regex for validation, just added one more rule to zod schema and error validation message ("Username can only contain alphanumerics and dashes.")

image


NiallJoeMaher (approved):
Nice! This is very helpful. 🥳

PR #811, opened 2024-03-08
refector:Add regex to username schema

John-Paul-Larkin:
# ✨ Codu Pull Request 💻

## Pull Request details

-Added regex to username field on additional details Zod schema, to bring it into line with the profile schema.

## Summary by CodeRabbit

- **Refactor**
	- Improved validation for user details:
		- `Username` now requires a minimum length and only allows alphanumeric characters and dashes.
		- `Location` now requires at least 1 character.
	- Added username normalization check and update in the `Page` function.


schema/additionalUserDetails.ts
```diff
@@ -14,8 +14,12 @@ export const slideOneSchema = z.object({
...
     .trim()
-    .min(3, "Min name length is 2 characters.")
-    .max(40, "Max name length is 40 characters."),
+    .min(3, "Min username length is 3 characters.")
+    .max(40, "Max username length is 40 characters.")
+    .regex(
+      /^[a-zA-Z0-9-]+$/,
+      "Username can only contain alphanumerics and dashes.",
+    ),
   location: z.string().min(1, "Location is required"),
```
JohnAllenTech, line 23: Worth updating this to something greater? Cant think of many locations with 1 character 🤣

John-Paul-Larkin: Well, the people of [Å in Norway](https://maps.app.goo.gl/5jSuqnHU4tez7j5E6) may be a little upset with that assumption! : 😊

The select is actually populated from an array of country names. The validation is just to ensure there is some value.

JohnAllenTech: I stand corrected. My apologies to the Norwegians  🤣 Ah ok this makes sense then

NiallJoeMaher (changes requested):
We also need to add fix for where it's getting passed or else a user won't be created:
https://github.com/codu-code/codu/blob/9a56909742b57d5fb48c2c2dabc352d5add85e74/app/(app)/settings/page.tsx#L42C5-L45C35
We can fix it here using: https://www.codu.co/articles/remove-accents-from-a-javascript-string-skgp1inb

John-Paul-Larkin:
> We also need to add fix for where it's getting passed or else a user won't be created: https://github.com/codu-code/codu/blob/9a56909742b57d5fb48c2c2dabc352d5add85e74/app/(app)/settings/page.tsx#L42C5-L45C35 We can fix it here using: https://www.codu.co/articles/remove-accents-from-a-javascript-string-skgp1inb

Yes, I read that article when you posted it.
Maybe I have this wrong, but I thought the cleaned string is required when the account is first created?
ie, at this stage the string will have already been cleaned?

You have not implemented the code in the article yet, as far as I can tell?
I will have a look this evening.

NiallJoeMaher:
> > We also need to add fix for where it's getting passed or else a user won't be created: https://github.com/codu-code/codu/blob/9a56909742b57d5fb48c2c2dabc352d5add85e74/app/(app)/settings/page.tsx#L42C5-L45C35 We can fix it here using: https://www.codu.co/articles/remove-accents-from-a-javascript-string-skgp1inb
> 
> Yes, I read that article when you posted it. Maybe I have this wrong, but I thought the cleaned string is required when the account is first created? ie, at this stage the string will have already been cleaned?
> 
> You have not implemented the code in the article yet, as far as I can tell? I will have a look this evening.

Correct! I haven't done it yet. How we create the user is redirecting to the settings page and if there is no user it auto-magically creates that user. Here's where that bug has been slipping through and needs cleaning.

John-Paul-Larkin:
> Correct! I haven't done it yet. How we create the user is redirecting to the settings page and if there is no user it auto-magically creates that user. Here's where that bug has been slipping through and needs cleaning.

I think this should do it. I tested it with special chars and it worked as intended.

My GitHub username is "john-paul-larkin"
However, when I am testing this the username field on account creation is always null. I am not sure why?

app/(app)/settings/page.tsx
```diff
@@ -30,13 +30,32 @@ export default async function Page() {
     redirect("/get-started");
   }
 
-  const user = await prisma.user.findUnique({
+  let user = await prisma.user.findUnique({
```
NiallJoeMaher: Any reason this is a let now?

John-Paul-Larkin: If the username string contains an accent we reassign user variable to the updated user on line 47.
> user = await prisma.user.update({

This updates user is then passed on to the settings page. 


Also coderabbits suggestion
> Error Handling: There's no explicit error handling around the database update operation. 

Should I wrap the update in a try.. catch?
The other prisma update in this component does not have one.

NiallJoeMaher: Aha! I get why I'm confused. 

The change for the user cleanup should be in this block:

```ts
 if (!user?.username) {
    const nanoid = customAlphabet("1234567890ABCDEFGHIJKLMNOPQRSTUVWXYZ", 3);
    const initialUsername = `${(session.user.name || "").replace(
      /\s/g,
      "-",
    )}-${nanoid()}`.toLowerCase();

    const newUser = await prisma.user.update({
      where: {
        id: session.user.id,
      },
      data: {
        username: initialUsername,
      },
      select,
    });

    return <Content profile={newUser} />;
  }
```

This is where the initial username is set. This is where the bug slips in (not above; that's just fetching the profile.) Your schema should stop updating the profile to a username with special characters then (which it already does).

John-Paul-Larkin: Ah... right. I also now get why I am confused :laughing: 

I was under the mistaken impression that the username field was coming for the GitHub account. Which is why I was asking this earlier.
>My GitHub username is "john-paul-larkin"
However, when I am testing this the username field on account creation is always null. I am not sure why?

I will fix it now.

NiallJoeMaher (approved):
🌮 LGTM! The name does come from GitHub but it is done at the auth layer. 🦾

PR #848, opened 2024-03-18
migrate to drizzle notification routes

John-Paul-Larkin:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details

I've refactored the following notification.ts routes:
- delete
- deleteAll
- getCount

The return value from delete and deleteAll routes is not used on the front end. Despite this, I followed the same return pattern as the Prisma queries I was replacing. ie
- delete : The id of the deleted record
- deleteAll : The count of deleted records.

The return value of getCount is used on the front end.

I will include the get route  in a seperate PR because it is more complicated.
 
## Any Breaking changes

- Hopefully not, but I would still like to test each route

## Summary by CodeRabbit

- **Refactor**
	- Updated database access method to improve performance.
	- Enhanced notification fetching and deleting logic for better efficiency.
	- Modified return values from delete operations to ensure accuracy.
	- Improved notification count operation for more reliable results.

### DIFF
--- a/server/api/router/notification.ts
+++ b/server/api/router/notification.ts
@@ -4,6 +4,8 @@ import {
   GetNotificationsSchema,
   DeleteNotificationSchema,
 } from "../../../schema/notification";
+import { notification } from "@/server/db/schema";
+import { count, eq } from "drizzle-orm";
 
 export const notificationRouter = createTRPCRouter({
   delete: protectedProcedure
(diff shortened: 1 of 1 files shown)


JohnAllenTech:
Nice 🚀 we are getting there with the migration! 

Great to see some of them are mapping 1:1 like this

NiallJoeMaher (approved):
🌮 Great job! Flying these days. 🛸

John-Paul-Larkin:
> Nice 🚀 we are getting there with the migration! 
> 
> Great to see some of them are mapping 1:1 like this 

Yes, getting there and I'm learning a lot in the process.
I will leave the hard ones to you 😀

PR #857, opened 2024-03-20
chore: migrated create comment function to drizzle

JohnAllenTech:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details

- Pretty much a lift and shift. Nothing to complex

## Any Breaking changes

- Potentially if we dont migrate the DB to match the Drizzle schema before we deploy this
- Drizzle schema has updatedAt as defaultNow() but it seems prisma had some type of @updatedAt which was doing this. Therefore theres a nonNull constraint on updatedAt from the prisma schema and a defaultNow() from the drizzle schema. To overcome this I was able to drop the DB and recreate it from the drizzle schema. We may not be able to do this in Prod

## Associated Screenshots
Showing the implemented comments

## Summary by CodeRabbit

- **Refactor**
	- Enhanced comment and notification handling for improved performance and consistency.

### DIFF
--- a/server/api/router/comment.ts
+++ b/server/api/router/comment.ts
@@ -11,6 +11,7 @@ import {
   NEW_COMMENT_ON_YOUR_POST,
   NEW_REPLY_TO_YOUR_COMMENT,
 } from "../../../utils/notifications";
+import { comment, notification } from "@/server/db/schema";
 
 export const commentRouter = createTRPCRouter({
   create: protectedProcedure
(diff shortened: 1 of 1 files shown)


NiallJoeMaher:
So just for clarity, what exactly could break when I deploy this? Or what do I need to change? We should be matching the schema of Prisma for the migration so that's why this worries me 😹 

Do we need to make sure the edit function is completed with this PR? Just so the updatedAt can be manually updated each time too?

JohnAllenTech:
> So just for clarity, what exactly could break when I deploy this? Or what do I need to change? We should be matching the schema of Prisma for the migration so that's why this worries me 😹
> 
> Do we need to make sure the edit function is completed with this PR? Just so the updatedAt can be manually updated each time too?

Sure so the drizzle comment schema is
image

and updatedAt is notNull and defaults to now() which is all good.

However the prisma schema doesnt have the default to now()
image

So if you initialise the DB with the prisma schema and then try to use the drizzle schema to update a comment. Drizzle will think its ok to leave updatedAt as null **but** Prisma/ Postgres will not be happy and will throw an error.

However if you initialise the DB with the Drizzle schema then everything will work as per. I didnt manage to get the migration working myself but I didnt put too much time into it

NiallJoeMaher:
> > So just for clarity, what exactly could break when I deploy this? Or what do I need to change? We should be matching the schema of Prisma for the migration so that's why this worries me 😹
> > Do we need to make sure the edit function is completed with this PR? Just so the updatedAt can be manually updated each time too?
> 
> Sure so the drizzle comment schema is image
> 
> and updatedAt is notNull and defaults to now() which is all good.
> 
> However the prisma schema doesnt have the default to now() image
> 
> So if you initialise the DB with the prisma schema and then try to use the drizzle schema to update a comment. Drizzle will think its ok to leave updatedAt as null **but** Prisma/ Postgres will not be happy and will throw an error.
> 
> However if you initialise the DB with the Drizzle schema then everything will work as per. I didnt manage to get the migration working myself but I didnt put too much time into it

I think this okay! Its just because Drizzle doesn't offer that auto update now function but we will always create an object with a default date if it's not provided at creation time. I think it should be fine. Now let's watch the world burn 😂

NiallJoeMaher (commented):
🌮 🌮 🌮 🌮 🌮 🌮

NiallJoeMaher (approved):
I meant to approve 😂 

🌮

JohnAllenTech:
@NiallJoeMaher i don't think we do... if you look at the PR we aren't providing the updatedAt property and are expecting the drizzle to sort it for us.

I can make this change in the follow on PR to ensure we always send one anyway but this might have broken the commment creation. I'll check dev

NiallJoeMaher:
Oh yeah just throw it into the next PR then and we will be sorted 😎

PR #864, opened 2024-03-24
Migrating next auth from Prisma to Drizzle

JohnAllenTech:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details

- Migrating next auth to use the drizzle adaptor.
- Drizzle adaptor is **very** opinionated about column names. It currently does not support anything but the default column names.
- Initially I tried to create a custom adaptor so we could use our existing DB schema but this was a bit of a nightmare.
- Now searching for banned users using drizzle

## Any Breaking changes

- If we merge it and dont do the schema migration it will break auth as Drizzle adaptor has different table names for users,sessions and verification tokens

## Associated Screenshots/Screenshares

https://github.com/codu-code/codu/assets/46611809/c66d8f53-a451-466f-9590-bbf39da64ea5

Heres Login via Github and PasswordLess working on local

## Summary by CodeRabbit

- **New Features**
	- Introduced a new database schema, defining tables and relationships for various entities like accounts, communities, posts, etc.
	- Added the `@auth/drizzle-adapter` package to enhance authentication processes.
- **Bug Fixes**
	- Updated authentication logic to correctly handle banned users.
- **Refactor**
	- Replaced `PrismaAdapter` with `DrizzleAdapter` in authentication server logic.
	- Database schema updates including renaming of tables and fields, changing data types, and modifying constraints.
- **Documentation**
	- Updated JSON schema definitions and journal entries to reflect database schema changes.


drizzle/meta/0001_snapshot.json
```diff
@@ -1404,11 +1404,6 @@
       "foreignKeys": {},
       "compositePrimaryKeys": {},
       "uniqueConstraints": {
-        "VerificationToken_identifier_unique": {
```
JohnAllenTech, line 1407: This is a weird phantom constraint that I cant see any other reference to. It might have been me doing some sketchy stuff earlier on in the migration but its unused and not needed so I have removed it from all the snapshots

The reason I removed it is because when I ask drizzle to generate a schema migration it looks at the snapshots and doesnt find this constraint in use anymore and trys to remove it. But it doesnt exist so this breaks the migration

server/db/schema.ts
```diff
@@ -10,79 +10,44 @@ import {
...
+  userId: text("userId")
+    .notNull()
+    .references(() => user.id, { onDelete: "cascade" }),
+  expires: timestamp("expires", { mode: "date" }).notNull(),
+});
 
-export const accountRelations = relations(account, ({ one, many }) => ({
-  user: one(user, { fields: [account.userId], references: [user.id] }),
-}));
+export const account = pgTable("account", {
```
JohnAllenTech, line 36: Copy and pasted these schemas from https://authjs.dev/getting-started/adapters/drizzle

server/db/schema.ts
```diff
@@ -133,22 +98,16 @@ export const tagRelations = relations(tag, ({ one, many }) => ({
...
-  },
-  (t) => {
-    return {
-      indx0: unique().on(t.identifier, t.token),
-    };
+    expires: timestamp("expires", { mode: "date" }).notNull(),
   },
+  (vt) => ({
+    compoundKey: primaryKey({ columns: [vt.identifier, vt.token] }),
+  }),
```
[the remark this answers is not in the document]
JohnAllenTech: Its used by next auth for passwordLess login. It happens in the background

NiallJoeMaher (approved):
🌮 Thanks for the helpful comments too!

PR #879, opened 2024-04-02
fix: fixed problem with update bookmarks

Maczi01:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details

Fixed problems with update bookmarks - after click icon, article will be added to bookmarks + toast "Bookmark updated"

## Any Breaking changes

Important. I am not sure it is most efficient solution. 

In onSuccess I am invalidating query, with queryClient.invalidateQueries(); 
Imo problem is with invalidate all queries, not particular one. I tried with:
queryClient.invalidateQueries([myBookmarks), but it doesn't work

## Associated Screenshots

image

## Summary by CodeRabbit

- **New Features**
	- Introduced success notifications for bookmark updates in article previews.
	- Refactored bookmark mutation handling to improve user experience.


NiallJoeMaher (changes requested):
Thanks for this! ❤️ 

Aha, the problem here is we need an optimistic update. Basically, in local development it's hard to test because everything happens fast. But it's so the bookmark instantly is filled or unfilled when bookmarking or unmarking. In production you'll notice a half second delay which could cause the user to hit it multiple times. This will behave the exact same except we have a toast now too. 

So the solution doesn't need a notification but it does need to look like the user has instantly bookmarked something (and then only alert them if something went wrong which should be rare).

Maczi01:
Whooo, optimistic update can be perfect here.
I am struggling with implement it, some query keys problem:

https://tanstack.com/query/v4/docs/framework/react/guides/optimistic-updates

Maybe we can change approach a littie bit and use disabled property on button? Something like:
 <button disabled={isLoading},
 
 I can get isLoading from useMutation.

NiallJoeMaher:
> Whooo, optimistic update can be perfect here. I am struggling with implement it, some query keys problem:
> 
> https://tanstack.com/query/v4/docs/framework/react/guides/optimistic-updates
> 
> Maybe we can change approach a littie bit and use disabled property on button? Something like: <button disabled={isLoading},
> 
> I can get isLoading from useMutation.

Sorry I'm slow getting back to you. Yeah something like that might be good! I'll also take a look and see what options there is

Maczi01:
@NiallJoeMaher can you check it now? button will be disable during action, it prevent sending another request. Should be helpful :-) If not, we can talk about better solution

NiallJoeMaher (changes requested):
This is awesome, and would work very well. I'm going to suggest something more opinionated because I like how it looks 🙈 

I was looking at the docs for "optimistic updates" and found this: https://tanstack.com/query/latest/docs/framework/react/guides/mutations#mutation-side-effects

```js
  onMutate: variables => {
    // A mutation is about to happen!
    // We can update right away
    // Return a context containing data to use when for example rolling back
    return { id: 1 }
  },
  onError: (error, variables, context) => {
    // An error happened!
    console.log(`rolling back optimistic update with id ${context.id}`)
  },
```

I'm not sure how the logic would work inside it but basically, on mutate, update the UI right away (as if the bookmark worked instantly) and then only on error roll back and present an error toast. I think a toast pop up for every bookmark success might get annoying especially on small screens. 



Example:
https://tanstack.com/query/v3/docs/framework/react/examples/optimistic-updates-typescript

PR #915, opened 2024-04-24
Fix: #914 disable like/bookmark buttons when request is pending

### LINKED ISSUE #914: Liking a post continuously causes the like count to to decrease
If you click fast on a posts like button it causes the like count to decrease

We need to disable the like button when it's in the process of liking/unliking a post

boldurean:
# ✨ Codu Pull Request 💻

Codu Logo

## Pull Request details
 
- Added disabled attribute to like/bookmark buttons while requests are pending.
- use isLoading (bool) prop instead of status as non any other mutation statuses are used in component.

## Any Breaking changes

- None

## Associated Screenshots

- 'None'

## Summary by CodeRabbit

- **Refactor**
	- Enhanced the like and bookmark functionalities in the Article Menu to improve performance and user feedback during loading states. Buttons are now disabled when actions are pending to prevent multiple submissions.


boldurean:
Ok, deleting branch didn't deleted reference from the issue. So will try do another one :O

PR #992, opened 2024-08-15
Refactor/onboarding form

John-Paul-Larkin:
- Rename the server actions to easily identify them as such.
- Refactor the onboarding form to use the Tailwind UI Catalyst components.
- All colors/styles have remained as the default, we can review these to bring them in line with the new designs.
- No logic has been changed.
- I only imported the required Catalyst components because the others were resulting in eslint errors when I tried to commit them.


NiallJoeMaher (changes requested):
The folder name for `/tailwind-catalyts` should be `/components` as we will adjust and amend these components. It also might confuse anyone who has no idea what that means 😂

John-Paul-Larkin:
> The folder name for `/tailwind-catalyts` should be `/components` as we will adjust and amend these components. It also might confuse anyone who has no idea what that means 😂

Yes, I was not sure what to do there. I wanted to differentiate those components. I will change it now.

NiallJoeMaher (approved):

PR #1007, opened 2024-09-09
New components

John-Paul-Larkin:
# ✨ Codu Pull Request 💻

- Disable husky pre commit linting. husky has not been removed, the command has been commented out.

- Remove Catalyst components from Alpha
- Add all Catalyst components to components>catalyst


.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
```
NiallJoeMaher, line 2: Delete this entire file and references to Husky if it's not being used 🤖

NiallJoeMaher (changes requested):

.husky/pre-commit
NiallJoeMaher: @John-Paul-Larkin

John-Paul-Larkin: Oh yeah. 
I didn't delete it because I thought there may be other uses for it?
I actually thought we might use it to run prettier on pre commit.

NiallJoeMaher: I don't see anywhere else we need it so get rid of it everywhere I guess.

NiallJoeMaher (approved):
🌮

PR #1045, opened 2024-09-30
feat: add gitlab oauth provider

### LINKED ISSUE #1013: Add an option for GitLab login
## Context

Currently, we allow users to Sign up/Sign in with GitHub. 
We would like to offer the option to sign in using GitLab.

## Expected Behavior

The Codú project uses [Next Auth](https://next-auth.js.org/) as its open source authentication library.

Docs for the [GitLab provider](https://next-auth.js.org/providers/gitlab#options)

## Screenshots

Current login screen

## Additional info

You can follow the pattern on the GitHubProvider which you can find at 
>server>auth.ts

Any questions, feel free to ask.

sleithdylan:
# ✨ Codu Pull Request 💻

Closes #1013 

## Pull Request details

- Added GitLab as an OAuth Provider
- Added a new button in `app/(app)/get-started/_client.tsx` for logging in with GitLab

## Any Breaking changes

- I noticed once you log into GitHub or GitLab for the first time, you can only log in to that provider going forward. If you try to log into the other provider after already logging into the other you will get the error, [OAuthAccountNotLinked](https://authjs.dev/reference/core/errors#oauthaccountnotlinked) and then button won't work.
    - Do you want to address this in another PR or this one? we would need to decide how you want to handle it, if you want to return an error on the login or try to let them login with both providers. @John-Paul-Larkin @NiallJoeMaher

## Associated Screenshots

https://github.com/user-attachments/assets/6a227dc2-14b6-4886-aa75-4c6ced45baeb

### DIFF
--- a/app/(app)/get-started/_client.tsx
+++ b/app/(app)/get-started/_client.tsx
@@ -5,7 +5,7 @@ import Image from "next/image";
 import Link from "next/link";
 import { signIn } from "next-auth/react";
 import { useSearchParams } from "next/navigation";
-import { useState } from "react";
+import React, { useState } from "react";
 import { LockClosedIcon } from "@heroicons/react/20/solid";
 
 const GetStarted: NextPage = () => {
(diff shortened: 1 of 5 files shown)


John-Paul-Larkin:
@sleithdylan 
Great work Dylan, thanks.

> I noticed once you log into GitHub or GitLab for the first time
That deserves its own PR I think, would you create an issue please. 
We will assign you and you will be halfway to Hactoberfest already 🥳

NiallJoeMaher:
Is this one ready for review @sleithdylan?

app/(app)/get-started/_client.tsx
```diff
@@ -5,7 +5,7 @@ import Image from "next/image";
 import Link from "next/link";
 import { signIn } from "next-auth/react";
 import { useSearchParams } from "next/navigation";
-import { useState } from "react";
+import React, { useState } from "react";
```
NiallJoeMaher: The `React` import is unused, so I think we can revert.

sleithdylan:
@John-Paul-Larkin Perfect I will do that later once I get the chance.

@NiallJoeMaher Not just yet, I need to just to fix something before I open the PR.

app/(app)/get-started/_client.tsx
sleithdylan: Hi @NiallJoeMaher so when I was in the file making changes I was getting an error in VSCode for the fragment that was added for the email input for the alpha, "'React' refers to a UMD global, but the current file is a module. Consider adding an import instead".

Adding the import resolved the error.

sleithdylan: I was researching and if you also change `"jsx": "preserve"` to `"jsx": "react-jsx"` in the **tsconfig.json**, it should also resolve the issue


[https://stackoverflow.com/a/72460283](https://stackoverflow.com/a/72460283)

sleithdylan:
@John-Paul-Larkin @NiallJoeMaher 

I was able to allow users to log in to the other provider after creating their account with one of the providers but if you read my OP that I edited I explain that it might not be the best option. If you want I can remove it and address it in another PR.

Also, there is another issue I have been having a problem with, when logging into GitLab I get redirected to another screen which shows "Redirecting", we could also address that in another PR also if you want. I tried to get it to not do it but couldn't get it to work right now.

server/auth.ts
```diff
@@ -42,6 +43,12 @@ export const authOptions: NextAuthOptions = {
     GitHubProvider({
       clientId: process.env.GITHUB_ID || "",
       clientSecret: process.env.GITHUB_SECRET || "",
+      allowDangerousEmailAccountLinking: true,
```
JohnAllenTech: Drive by comment but if you add this it will be pretty easy to hijack someones account on Codu.

All I need to know is your GH email (which is in your commits) and the fact you dont have a gitlab account

Now that I think of it... whats the danger here? I was thinking of creating a Gitlab account with your email but I assume gitlab has an email verification step?

JohnAllenTech: image

Looks like gitlab has email verification before I can login correctly.

sleithdylan: @JohnAllenTech Yeah from my main post I explain that there is a security issue and it may not be the best way, @John-Paul-Larkin said I could address the issue of not being able to log into both providers in another issue so I might just do that, I am going to revert the changes and do that instead of using this "hack"

NiallJoeMaher: Yeah, agreed. Linking is a tricky one! So I think for the moment we can treat them as separate accounts but not allow signups with duplicate email accounts. 

A user can always use their email address to login with @JohnAllenTech's email login anyway. It's been outstanding since last year so we just need to plugin the onboarding flow now and connect all of that.

NiallJoeMaher (approved):
🌮 Good work! Thanks!

PR #1055, opened 2024-10-03
Debouncing liking a post in the hackiest way imaginable

### LINKED ISSUE #914: Liking a post continuously causes the like count to to decrease

JohnAllenTech:
# ✨ Codu Pull Request 💻

Fixes #914

## Pull Request details

- This is a disgusting way to do this and I wouldnt approve this PR
- I have tried to debounce useMutation 6 different ways but I cant get it to work. This is 100% a skill issue in terms of FE work.
- I hope someone sees this and is so disgusted they fix it correctly

## Any Breaking changes

- None

## Associated Screenshots

https://github.com/user-attachments/assets/103b693a-9547-4e4a-b6ea-fcfa92139d10

I am spam clicking the like button. Not you need to throttle your connection locally as we need to imitate real life network calls. Calls locally resolve to fast to replicate this bug without the throttling.


server/api/router/post.ts
```diff
@@ -218,13 +218,16 @@ export const postRouter = createTRPCRouter({
...
-              );
-            await tx
-              .update(post)
-              .set({
-                likes: decrement(post.likes),
-              })
-              .where(eq(post.id, postId));
+              )
+              .returning();
+            if (res.length !== 0) {
```
JohnAllenTech, line 223: Essentially we are ensuring there is a like to remove before be decrement the like count stored on the post table.

Previously even if no like was found we decremented the like count of the post. This lead to the ability to have negative likes

NiallJoeMaher: I think this should work. Could the same happen in the positive way (eg giving infinite likes?)

JohnAllenTech: No because theres a foreign key constraint that prevents a user creating two 'like' rows for the same post. This causes the transaction to fail.

For some reason when you try to delete something that doesnt exist it doesnt throw an error and instead just goes 👍🏻

NiallJoeMaher (approved):
🌮 Works until we find out how we are supposed to do it 😹

PR #1065, opened 2024-10-05
Add feature flag example

NiallJoeMaher:
# ✨ Codu Pull Request 💻

## Pull Request details

- Because we have hacky feature flags I wanted to create something a little more stable
- We are now using Posthog for feature flags
- This page can be given in issues to show an example of adding a feature flag to a page


NiallJoeMaher:
@John-Paul-Larkin added feature flags configuration so might be one to look at

PR #1086, opened 2024-10-07
feat: init e2e workflow setup

pkspyder007:
# ✨ Codu Pull Request 💻

Fixes #(issue)
- NA

## Pull Request details

- Created a workflow to run e2e tests under e2e env with pre-seeded db and github app setup 
- Requires few secrets to to added for db and github app setup

## Any Breaking changes

- None

## Associated Screenshots

- None


pkspyder007:
The tests will start passing once [this PR ](https://github.com/codu-code/codu/pull/1085)merged and rebased into the e2e workflow branched

pkspyder007:
@NiallJoeMaher @John-Paul-Larkin 
I am not having access to add secrets to the repo, if any of you could create a new supabase instance and add the 
`E2E_DATABASE_URL` in the secret we'll have a persistent E2E DB

Also we need to create github oauth client for testing github auth flows
That will require similar secrets please refer to the workflow implementaion

NiallJoeMaher:
@pkspyder007 - Do you think we should use an e2e database url or just give it a database URL it can use instead (rather than having a separate one). Eg. if we used `DATABASE_URL` would there be any issue?

pkspyder007:
> @pkspyder007 - Do you think we should use an e2e database url or just give it a database URL it can use instead (rather than having a separate one). Eg. if we used `DATABASE_URL` would there be any issue?

@NiallJoeMaher The current setup uses `DATABASE_URL` only I have just named secret to have E2E in the name so its clear to understand what this DB url exactly is.
Our `E2E_DATABASE_URL` will contain link to a postgres db with seeded data.

pkspyder007:
Also wanna add that running a docker image will utilise more resource and cost us more and lead to increased the CI times.

praveen-quizizz:
@NiallJoeMaher have a look whenever you get time

few action items
- check for login test flakiness
- fix seeding
- updating and revamping skipped tests
- move tests results to report on PR and host traces to something like S3

NiallJoeMaher (approved):
🌮

PR #1099, opened 2024-10-09
[FIX] API calls for Recent Bookmarks Sidebar

### LINKED ISSUE #1097: Add Error Boundary and Server-Side API Calls for Recent Bookmarks Sidebar
## Context

The Recent Bookmarks sidebar currently crashes the entire page if the API call fails. We need to improve the error handling so that the entire page is not affected by an API failure related to this component. 

## Expected Behavior

- Add an error boundary specifically around the SideBarSavedPosts component to handle failures gracefully.
- Ensure error handling applies to individual pieces of the component, so it does not affect the rest of the page.

## Screenshots

Home page with recent bookmarks

HarshitT00:
# ✨ Codu Pull Request 💻

Fixes #(issue)

## Pull Request details

Getting data and status in same query which is previously done saperately.

### DIFF
--- a/components/SideBar/SideBarSavedPosts.tsx
+++ b/components/SideBar/SideBarSavedPosts.tsx
@@ -6,14 +6,13 @@ import SideBarSavedArticlePreview from "./SideBarSavedArticlePreview";
 import Link from "next/link";
 
 export default React.memo(function SideBarSavedPosts() {
-  let { data: bookmarks } = api.post.myBookmarks.useQuery();
-  const { status: bookmarkStatus } = api.post.myBookmarks.useQuery();
+  const { data: bookmarks, status } = api.post.myBookmarks.useQuery();
 
   const howManySavedToShow = 3;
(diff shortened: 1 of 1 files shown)


components/SideBar/SideBarSavedPosts.tsx
```diff
@@ -6,45 +6,43 @@ import SideBarSavedArticlePreview from "./SideBarSavedArticlePreview";
 import Link from "next/link";
 
 export default React.memo(function SideBarSavedPosts() {
-  let { data: bookmarks } = api.post.myBookmarks.useQuery();
-  const { status: bookmarkStatus } = api.post.myBookmarks.useQuery();
+  const { data: bookmarks, status } = api.post.myBookmarks.useQuery();
```
NiallJoeMaher, line 9: I think this line alone should be good! Maybe we can pass the amount of bookmarks to the useQuery() so that we only get the right amount back rather than doing this on the frontend?

PR #1109, opened 2024-10-11
Settings screen update / Fix Drizzle Kit Studio

NiallJoeMaher:
A few fun changes here.

- Made the new settings screen responsive so it can be used as a template for other forms. 
- Added indexes to speed up fetching of posts and user profiles
- Fixed some schema issues that gets drizzle kit studio working again so you can inspect data
- Refactored the email verification system to ensure it can't be spammed and to stop it locking up emails

New settings screen on desktop
New settings screen on mobile


John-Paul-Larkin (dismissed):
Fabulous!

PR #1112, opened 2024-10-12
Fixes Refetch of SideBar Posts on Bookmarked

### LINKED ISSUE #834: Recent bookmarks - auto update UI and update feedback
## Expected Behavior

Please describe the behavior you are expecting

- When I click on the bookmark icon the "recent bookmarks" should auto add or remove the bookmark without having to refresh the page.

- When adding or removing a bookmarked article if the action is unsuccessful there should be a toast notification with simple actions to resolve.

## Current Behavior

What is the current behavior?

When I click on the bookmark icon in an article, the icon will change colour (provides feedback), 
but I need to manually refresh the page to see the updated bookmarks,

revanthnetha:
# ✨ Codu Pull Request 💻

Fixes #834

Approach:
- So whenever in article something is bookmarked or unbookmarked, refetch of the myBookmarks i.e.SidePosts would happen.
- In order to do this, when mutate post in Article Preview is successfull, I have called the myBookmarks Query, So  the recent data shown in SideBarPosts.
- **Refetch Syncs Data**: When refetch() is called in Component 1 (e.g., after a bookmark mutation), Component 2 will also display the most recent data because React Query syncs the cache automatically across components.

Todo :
-  Instead of slicing the SidePosts in FE. I have modified the Trpc Query to take the limit and render the limit bookmarks SidePosts.
- we can also do in way by making two queries one for fetching the bookmarks until the limit and other for total Count. But I have found making oneQuery to fetch all bookmarks and then slicing is optimized as this query : ctx.db.query.bookmark.count(..) is not supporting.

Happy to do changes if needed :)

## Any Breaking changes
- None

## Associated Screenshots
- None


components/SideBar/SideBarSavedPosts.tsx
```diff
@@ -6,14 +6,14 @@ import SideBarSavedArticlePreview from "./SideBarSavedArticlePreview";
 import Link from "next/link";
 
 export default React.memo(function SideBarSavedPosts() {
-  let { data: bookmarks } = api.post.myBookmarks.useQuery();
-  const { status: bookmarkStatus } = api.post.myBookmarks.useQuery();
+  const { data: bookmarksData } = api.post.myBookmarks.useQuery({
+    limit: 3,
```
John-Paul-Larkin: Instead of the magic number 3 for the limit, you should use the howManySavedToShow const

John-Paul-Larkin:
Hey @revanthnetha 
This works great, thanks.
I just left a small comment, but otherwise this is good to go 👍

components/SideBar/SideBarSavedPosts.tsx
```diff
@@ -6,14 +6,14 @@ import SideBarSavedArticlePreview from "./SideBarSavedArticlePreview";
...
+  });
+  const { status: bookmarkStatus } = api.post.myBookmarks.useQuery({});
```
John-Paul-Larkin: Can you please combine this query with the other one. This means we are making 2 queries when only one is necessary.

John-Paul-Larkin (changes requested):

revanthnetha:
Got it @John-Paul-Larkin . I'll make the changes as necessary.

revanthnetha:
Review the changes @John-Paul-Larkin . Happy to do if any changes needed :)


Could you please add the labels hacktober-fest and hactober-fest-accepted tags and add my name in assigness. It really means a lot.

John-Paul-Larkin:
Hey @revanthnetha 
This is great work. Thanks. 
I was playing with this earlier and I received the error Toast, but I dont know why that happened and try as I may I cant seem to recreate it. 

@NiallJoeMaher Do you think we remove the toast on error?
Apart from that, I think this is ready to go.

John-Paul-Larkin (approved):

Issue #998, opened 2024-09-03
.gifs don't work on image upload once processed


NiallJoeMaher:
## Context

The initial preview of a gif shows as expected but after it is processed by the lambda it loses the motion. 

This is because during the resize the gif is losing its format. 

## Expected Behavior

gifs can be resized but should still work. 

Code can be found in `/cdk/lambdas/uploadResize`

JohnAllenTech:
Oh wow.... my gif people need me!

Can you document the flow of how an image gets resized @NiallJoeMaher  or @John-Paul-Larkin ?

Is it

1. User adds image to an article
2. Image gets sent to lambda to be resized ?
3. If image is a gif its now broken?

NiallJoeMaher:
Yes, you'll see it in the CDK lambdas folder. From memory it's just because we convert those gifs to a static image

PR #1143, opened 2024-10-17
Lambda to allow gif types

NiallJoeMaher:
# ✨ Codu Pull Request 💻

Fixs #998 

## Pull Request details

- Add a conditional check for gifs to ensure they aren't converted to webp
- General cleanup of avatar resize lambda

## Any Breaking changes

- None


JohnAllenTech:
Nice @NiallJoeMaher 

Does this mean we can have gifs for avatars now?

NiallJoeMaher:
> Nice @NiallJoeMaher 
> 
> Does this mean we can have gifs for avatars now? 

I didn't enable it on this one it was for articles and discussions. But... We can get that working easily with the same logic 😂

PR #1155, opened 2024-10-19
Implemented Company Logo upload functionality

### LINKED ISSUE #1148: Image Upload Functionality Not Implemented for Company Logo in Jobs Create page
## Context

The image upload feature for the company logo is currently not functioning, as the implementation is missing.

## Expected Behavior

Users should be able to upload a company logo from their file explorer, and the selected image should be reflected in the UI after submission.

## Current Behavior

The company logo is not being updated after selecting an image from the file explorer. No upload occurs as the functionality is not yet implemented.

Nil2000:
# ✨ Codu Pull Request 💻

Fixes #1148 

## Pull Request details

- Implementation of Company Logo upload functionality.

## Any Breaking changes

- None


NiallJoeMaher:
Sorry about the delay I was on holidays 🌞 

The issue is for this part of the application, a person does not have to be authenticated, so we may need another safe action that works unauthenticated that we can pass the schema to. 

There is some solid work here! I'm hoping we can do a couple of updates if you don't mind. Would you be able to do it with actions instead of trpc? 

Something like here: https://github.com/codu-code/codu/blob/develop/app/actions/getUploadUrl.ts

You could make the new action template here: https://github.com/codu-code/codu/blob/develop/server/lib/safeAction.ts
Called `noAuthSafeAction`.

You can see it being imported and then being used here: https://github.com/codu-code/codu/blob/e4102dd9c421c10604ba07b5ffab11d3802cf3c7/app/(editor)/create/%5B%5B...paramsArr%5D%5D/_client.tsx#L37

NiallJoeMaher (changes requested):
Added a comment that you might have missed.

PR #1159, opened 2024-10-20
Feat/e2e testing for article sorting

John-Paul-Larkin:
# ✨ Codu Pull Request 💻

## Pull Request details

- Add Playwright tests for sorting on the article page
- Run test for three sorting modes -  "Newest", "Oldest" and "Top"

- Like count added to ArticlePreview
- I was required to pass the like count to the component, so If it non-zero, I display it on the Article preview with the heart icon. 
- I also added the count as a data attribute, so we can easily remove this if you dont like how it looks.

Article preview with like count

## Any Breaking changes

- None


components/ArticlePreview/ArticlePreview.tsx
```diff
@@ -124,6 +127,15 @@ const ArticlePreview: NextPage<Props> = ({
                 <>
                   <span aria-hidden="true">&middot;</span>
                   <span>{readTime} min read</span>
+                  {likes && (
```
JohnAllenTech, line 131: Oh interesting because you need a way to know how many likes an article has without clicking into it. That makes sense

e2e/articles.spec.ts
```diff
@@ -220,4 +220,52 @@ test.describe("Authenticated Articles Page", () => {
 
     await expect(page.getByText(commentContent)).toBeVisible();
   });
+
+  test.describe("Article sort order on article page", () => {
+    const sortOptions = ["newest", "oldest", "top"];
+
+    for (const sortOption of sortOptions) {
```
JohnAllenTech: Id recommend breaking ["newest", "oldest", "top"] into their own tests.

Yes we are repeating outselves but we also get test isolation ie. currently if this test breaks we will need to go into the actual test to see which option it breaks on

e2e/articles.spec.ts
```diff
@@ -220,4 +220,52 @@ test.describe("Authenticated Articles Page", () => {
...
+            date: article.querySelector("time")?.dateTime,
+            likes: parseInt(
+              article
+                .querySelector("[data-likes]")
+                ?.getAttribute("data-likes") || "0",
+            ),
+          }));
+        });
+
+        if (sortOption === "newest") {
```
JohnAllenTech: Comment above would clean up all this.

JohnAllenTech (commented):
Just a few comments. Overall it LGTM

e2e/articles.spec.ts
John-Paul-Larkin: Yeah, I was not sure what to do there. I will split it into separate tests. Thank John.

BTW is using the data attribute(for the like count ) the correct way to go about this?

JohnAllenTech: If your free to sync for 5 mins I can show you the way I get the most specific way to target an element.

TLDR if your not free. 

- Install playwright plugin for vs code
- Open the plugin view in vs code
- Use the pick locator function and hover over the element you want to target
- The plugin will give you the best locator to use

John-Paul-Larkin: @JohnAllenTech 
Great, I just hacked my way through writing a test today, without reading the docs 🫢 
I just downloaded that plugin, so will give it a try now. 

That would be greta. I am online if you are free?

NiallJoeMaher:
@JohnAllenTech, this one good to go?

e2e/articles.spec.ts
```diff
@@ -220,4 +220,58 @@ test.describe("Authenticated Articles Page", () => {
 
     await expect(page.getByText(commentContent)).toBeVisible();
   });
+
+  test("Should sort articles by Newest", async ({ page }) => {
```
JohnAllenTech: Do we want to copy and paste these tests into the unauthenticated view?

Does it need to be in both or just unauthenticated?

Given I believe the functionality of sorting is unaffected by a users auth status I think it only needs to be in unauthenticated

JohnAllenTech (commented):
Just a small comment. Tests LGTM

e2e/articles.spec.ts
John-Paul-Larkin: hm... Yes, I guess it shouldnt matter really?
I will copy it to unauthenticated.

components/ArticlePreview/ArticlePreview.tsx
NiallJoeMaher: Why do we need to know the amount of likes before it's loaded? 🤔 I don't think it's shown on the card?

John-Paul-Larkin: @NiallJoeMaher 
> Why do we need to know the amount of likes before it's loaded? 🤔 I don't think it's shown on the card?

I added it to the card in this PR. 
if you take a look at the PR description you will see how it looks. 
Heart icon with the like count. If there are no likes it wont show.

I figured you wouldnt like this 🤣  so I added it as a data attribute and can edit to remove it from the card.
The like count has to be included somewhere in the html, so playwright can grab it. For the sort by top test.

NiallJoeMaher (approved):

PR #1172, opened 2024-10-24
Chore/remove styles from globals.css

John-Paul-Larkin:
# ✨ Codu Pull Request 💻

## Pull Request details

This PR introduces a new Higher-Order Component (Focusable) to replace the use of the focus-style class, as part of a transition away from globals.css and heavy reliance on Tailwind's @apply directive.

The Focusable HOC wraps relevant components and applies focus styles. During implementation, it was observed that certain elements were missing focus classes, even though they likely should have them, and there is conflicting focus styles in some cases. These issues will be dealt with in a seperate PR.

Key Changes:
Removal of .focus-style class: The Focusable HOC now dynamically applies focus styles using React’s cloneElement function. While the React documentation mentions that cloneElement is uncommon and can lead to fragile code, I think it is appropriate for this simple use case. 
The alternative would involve creating separate HOCs for each element type (e.g., button, a, Link), which may lead to unnecessary complexity.

Next Steps:
If this approach is approved, the same pattern can be used to further transition away from @apply styles and reduce dependency on globals.css.

## Sperate to the HOC

Inlined Single-Use Classes: The following classes, previously only used in a single location, have had their styles embedded directly into the components. Also, the single use CSS variables have been removed.
.twitter
.github
.discord
.youtube

Commented Out Unused Styles for Review:
Temporarily commented out .rehype-code-title and .highlight-line styles, pending confirmation that they are no longer needed.

## Any Breaking changes

None

### DIFF
--- a/app/(editor)/create/[[...paramsArr]]/_client.tsx
+++ b/app/(editor)/create/[[...paramsArr]]/_client.tsx
@@ -557,7 +557,7 @@ const Create = ({ session }: { session: Session | null }) => {
                                 type="button"
                                 className="relative flex w-full focus:outline-none focus:ring-2 focus:ring-pink-300 focus:ring-offset-2 active:hover:bg-neutral-50 disabled:opacity-50"
                               >
-                                <div className="input-base flex max-w-full flex-1 overflow-hidden border text-left">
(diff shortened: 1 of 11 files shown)


John-Paul-Larkin:
@NiallJoeMaher 
We should discuss a plan for this, since im not sure if i am going about this correctly.

PR #1178, opened 2024-10-25
Fix callback getting appended to the url when signin button is clicked on login page

### LINKED ISSUE #777: Appending callback to url when hitting Sign in button on Nav
## Context

If you spam click the sign in button it keeps appending 'www.codu.co%2Fget-started%3FcallbackUrl%3Dhttps%253A%252F%252Fwww.codu.co%252F' to the url.
image

## Expected Behavior

The url would update on first click and would not update on subsequent clciks

What is the current behavior?

The url is appended with 'www.codu.co%2Fget-started%3FcallbackUrl%3Dhttps%253A%252F%252Fwww.codu.co%252F' each time.

## Steps to reproduce

1. Navigate to https://www.codu.co/
2. Click Sign in button in Nav
3. Click Sign in button in Nav again
4. Look at the url

## Additional info

No loss in functionality as far as I can tell so maybe not a big deal

IdrisGit:
# ✨ Codu Pull Request 💻

Fixes #777 

## Pull Request details

- NextAuth.js `singin()` method redirects to the login page and appends the current page URL to the callback.
- Added guard clause to prevent the navbar sign in button from calling the `signin()` method.

## Any Breaking changes

- None

## Associated Screenshots

- None

### DIFF
--- a/components/ArticleMenu/ArticleMenu.tsx
+++ b/components/ArticleMenu/ArticleMenu.tsx
@@ -139,7 +139,7 @@ const ArticleMenu = ({
               }}
             >
               <HeartIcon
-                className={`w-6 h-6${
+                className={`h-6 w-6 ${
                   data?.currentUserLiked
                     ? "fill-red-400"
                     : "fill-neutral-400 dark:fill-neutral-600"
@@ -162,7 +162,7 @@ const ArticleMenu = ({
             }}
           >
             <BookmarkIcon
-              className={`w-6 h-6${
+              className={`h-6 w-6 ${
                 data?.currentUserBookmarked
                   ? "fill-blue-400"
                   : "fill-neutral-400 dark:fill-neutral-600"
--- a/components/Nav/Nav.tsx
+++ b/components/Nav/Nav.tsx
@@ -1,5 +1,6 @@
 "use client";
 import { api } from "@/server/trpc/react";
+import { usePathname } from "next/navigation";
 import {
   Disclosure,
   DisclosureButton,
@@ -40,6 +41,8 @@ const Nav = ({
     enabled: session ? true : false,
   });
 
+  const pathname = usePathname();
+
   const userNavigation = [
     {
       name: "Your Profile",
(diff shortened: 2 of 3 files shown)


components/ArticleMenu/ArticleMenu.tsx
```diff
@@ -139,7 +139,7 @@ const ArticleMenu = ({
               }}
             >
               <HeartIcon
-                className={`w-6 h-6${
```
JohnAllenTech: How come you needed to change the order of these?

components/Nav/Nav.tsx
```diff
@@ -55,6 +58,18 @@ const Nav = ({
 
   const hasNotifications = !!count && count > 0;
 
+  const handleSignInPageNavigation = () => {
```
JohnAllenTech, line 61: Nice! This is very much a minor issue but it annoyed me and means we need to add guards in E2E tests. This will remove that need

JohnAllenTech (commented):
Just some questions around the class property swap. Seems that its not needed to me

components/ArticleMenu/ArticleMenu.tsx
IdrisGit: This is part of an already merged change, I had done a git pull from upstream after committing my changes and since I had global rebase to true, it caused this commit to be moved ahead of my commit.

[Reference to the merged commit](https://github.com/codu-code/codu/commit/871253f2f73106429f40e28fac42f24ba3a04c97)

components/Nav/Nav.tsx
IdrisGit: Thanks, I am happy to help :)

Helping maintainers to solve these minor but time consuming issues is my goal for this Hacktoberfest.

components/ArticleMenu/ArticleMenu.tsx
JohnAllenTech: Ok this makes sense. Thanks for the clarification

NiallJoeMaher (approved):

Issue #1168, opened 2024-10-22
E2E test - /settings page - Multiple issues available


John-Paul-Larkin:
## Context

We need to ensure that our core application features and user flows are fully covered by end-to-end (E2E) tests. We are using Playwright for these tests, and the workflow framework for E2E testing is already in place.

If you navigate to the e2e folder in the root of the project, you'll find that @JohnAllenTech has written numerous tests for the article page. These tests have been grouped into two categories: authenticated and unauthenticated. Please ensure that this pattern is followed when adding new tests.

## Expected Behavior

Below is a list of possible tests for the `/settings` page. These are suggested tests generated with the help of ChatGPT, meant to provide a ***starting point***. You should use your discretion to refine these tests and account for relevant edge cases. Ensure that tests for each input field align with the current Zod validation schema. 

Additionally, feel free to propose other valid tests that might apply specifically to the functionality of the settings page.

Tests should be standalone and focus on a single purpose, ensuring each test verifies only one specific behavior or functionality.

Test should be added to e2e/settings.spec.ts

Please leave a comment specifying the tests you plan to write, so others don't duplicate the same efforts. 
All PRs are accepted for Hacktoberfest.

**Note:** For this page, the unauthenticated test is only necessary to verify that the user is correctly redirected to the `/get-started` page.

**Note:** As multiple people may contribute to this issue, please ensure you pull the latest changes from the upstream and resolve any conflicts. [Sync the upstream](https://www.codu.co/articles/effortless-upstream-syncing-a-custom-alias-for-git-users-fowfliwo)

## Tests

1. Profile Picture Upload
Test: Upload a valid profile picture (JPG, PNG, or GIF).
Scenario: Simulate selecting an image file, uploading it, and verifying that it is displayed correctly in the UI.
Edge Case: Test for an invalid file format (e.g., PDF, BMP) and ensure proper error messaging.
2. Full Name Update
Test: Update the full name field and save the changes.
Scenario: Input a new name, click "Save Changes", and verify that the new name is updated and persisted across sessions.
Edge Case: Test for maximum length or invalid characters, if there are any validation rules.
3. Username Update
Test: Update the username field and save the changes.
Scenario: Input a new username, click "Save Changes", and verify that the username is reflected across the platform.
Edge Case: Ensure that no two users can have the same username and that errors are displayed for duplicates.
4. Bio Field Update
Test: Update the bio field with valid text.
Scenario: Input a bio within the character limit, save, and verify that it updates correctly.
[...]

PR #1179, opened 2024-10-26
fixed errors on `/app/articles/app/page.tsx` and added types for isomorphic-dompurify

petercr:
# ✨ Codu Pull Request 💻

Fixes 
It is related to issue #1168 and E2E testing, but more importantly it fixes an issue on the articles page which would not load the corresponding article without crashing.

## Pull Request details
- I tried to view /articles and then click the corresponding link.
- The Next web server would crash with error: `can't resolve module @tiptap/html`
- I fixed the import issues with tiptap/html
- I then added the types for the isomorphic-dompurify package
- Manual testing confirms the feature is now working. 

## Any Breaking changes

None.

## Associated Screenshots

Screenshot of the issue now resolved with current PR:
Screenshot_26-10-2024_121041_localhost


app/(app)/articles/[slug]/page.tsx
```diff
@@ -119,11 +119,15 @@ const ArticlePage = async ({ params }: Props) => {
 
           {isTiptapContent ? (
             <div
-              dangerouslySetInnerHTML={{ __html: renderedContent }}
+              dangerouslySetInnerHTML={{ __html: renderedContent ?? "" }}
```
Comment (NiallJoeMaher) [inline app/(app)/articles/[slug]/page.tsx]: If we fallback to `""` should we not have an error thrown?

Comment (petercr) [reply app/(app)/articles/[slug]/page.tsx]: So without the extra `|| ""` you get a Typescript error warning you that if it ends up getting passes a null value, there will be problems.

It seems like an easy fix, but if you wanted to default to something else I will be happy to change it 👍🏻

Or if you want me to pass in or throw a new error I could add that too.

Comment (NiallJoeMaher) [reply app/(app)/articles/[slug]/page.tsx]: I think if there is no content it should probably be notFound since there is no data (or content) for that page? That make sense?

Comment (petercr) [reply app/(app)/articles/[slug]/page.tsx]: Great I will just bring in the `NotFound` component then, and replace the empty string instance with that 👍

NiallJoeMaher (approved):
🌮 🦖 Nice one!

Issue #468, opened 2023-10-02
E2E Testing


NiallJoeMaher:
## Context

Right now, we have no e2e testing but Playwright setup.
We should have end-to-end testing enabled to make sure people are more comfortable changing things without fear of breaking things. 

The first thing would be some input on the appropriate testing.

Then, once we have decided on some appropriate areas, we can split the testing into issues to share the workload. 

## Expected Behavior

Test all major functionality for the application in e2e tests. 

## Current Behavior

No testing 😢

JohnAllenTech:
Working on this one over the weekend @NiallJoeMaher @John-Paul-Larkin 

Any particular tests you would like to see added?

I have the GH auth flow working in playwright so we can test authenticated portions of the application now like Creating a post, commenting on a post etc

John-Paul-Larkin:
Hey @JohnAllenTech 

I have not done any testing, so I am eager to learn. I was reluctant to make issues until I have a better understanding myself.

> Any particular tests you would like to see added?

Long term, we should cover as much as possible, but for now the plan is to break this up in to specific issue.
Auth flow is a great start, and since @sleithdylan recently added GitLab login, we should also add tests for that. 

- Tests for the main forms, which are the user settings page and the onboarding form.
- Tests for the alpha editor, which will replace the current editor shortly.
- Navigation tests between the key pages.
- Search functionality tests

Im really just guessing here, but I can look into it properly later and we can make a proper plan on Mondays livestream? @NiallJoeMaher

NiallJoeMaher:
Yes @JohnAllenTech. We can make a heap of issues from this so we can get contributors helping out too. 
- Confirm all the pages work as expected with/without auth.
- Confirm all links work (mobile/desktop) with/without auth.
- Confirm features like comments, likes, bookmarks, creating posts etc.
But yeah, it's basically test the f*ck out of everything!

PR #1181, opened 2024-10-26
chore: moving out user auth out into a util for E2E tests

JohnAllenTech:
# ✨ Codu Pull Request 💻

Helps with #1168 and #468 

## Pull Request details

So eventually we will need multiple user personas in a single E2E test.

One example is a user replying to another users comment. This PR is a first step to support multiple users in our E2E test. Follow on PRs will create a new loggedInAsUserTwo function which will be used to reply to 'user on'

I have also switched from having our tests run as authenticated by default

## Any Breaking changes

- None

## Associated Screenshots

- None


NiallJoeMaher (dismissed):

JohnAllenTech:
Something strange happening when deployed. Will do some troubleshooting. Everything is working locally with the same commands.

JohnAllenTech:
@NiallJoeMaher it seems that the PR workflow always runs from the main branches version.

I have a new env var thats needed for the tests to pass. Not sure how to work around this..

I am pretty confident it will pass the E2E tests if you merge it though. Can we see and revert if thats not the case?

JohnAllenTech:
image
Heres the reason its failing.

It needs E2E_USER_ONE_SESSION_ID which i have added to the E2E test workflow file.

NiallJoeMaher (approved):
YOLO

JohnAllenTech:
It worked anyway! I imagine this will be the same for the other two... maybe

PR #1182, opened 2024-10-26
Feat/multiple user support in e2e tests

JohnAllenTech:
# ✨ Codu Pull Request 💻

Will need to be merged after #1181

## Pull Request details

Follow on from #1181. Now creating two E2E users and supporting using them in our E2E test suite.

This unlocks the ability to test interaction between two users in an E2E Test. ie one user comments on another users comment

## Any Breaking changes

- None, however users will need to reseed their DBs locally to get the tests to pass again.

You can do this by running

```
npm run db:nuke
npm run db:migrate
npm run db:seed
```

## Associated Screenshots

image

All tests passing locally

PR #1183, opened 2024-10-26
Feat/adding test for replying to a comment

JohnAllenTech:
# ✨ Codu Pull Request 💻

Will need to be merged after #1182 

This closes out the multiple E2E user work

## Pull Request details

This PR introduces a new test to check a user can reply to another users comment.
In order to isolate this test, we need to prepopulate the DB with an article and a comment already made on it. This is done by the global setup file for the E2E test suite.

## Any Breaking changes

- None

## Associated Screenshots

image

image


JohnAllenTech:
This is the same job @NiallJoeMaher it wont work because its using the main branches e2e workflow file but the tests require a new env variable.

If its merged I am reasonably sure it will work

NiallJoeMaher (approved):

Issue #1169, opened 2024-10-22
E2E test - /my-posts page - Multiple issues available


John-Paul-Larkin:
## Context

We need to ensure that our core application features and user flows are fully covered by end-to-end (E2E) tests. We are using Playwright for these tests, and the workflow framework for E2E testing is already in place.

If you navigate to the e2e folder in the root of the project, you'll find that @JohnAllenTech has written numerous tests for the article page. These tests have been grouped into two categories: authenticated and unauthenticated. Please ensure that this pattern is followed when adding new tests. 

## Expected Behavior

Below is a list of possible tests for the `/my-posts` page. These are suggested tests generated with the help of ChatGPT, meant to provide a ***starting point***. You should use your discretion to refine these tests and account for relevant edge cases. 

Additionally, feel free to propose other valid tests that might apply specifically to the functionality of the '/my-posts' page.

Tests should be standalone and focus on a single purpose, ensuring each test verifies only one specific behavior or functionality.

Test should be added to e2e/my-posts.spec.ts

Please leave a comment specifying the tests you plan to write, so others don't duplicate the same efforts. 
All PRs are accepted for Hacktoberfest.

**Note:** For this page, the unauthenticated test is only necessary to verify that the user is correctly redirected to the `/get-started` page.

**Note:** As multiple people may contribute to this issue, please ensure you pull the latest changes from the upstream and resolve any conflicts. [Sync the upstream](https://www.codu.co/articles/effortless-upstream-syncing-a-custom-alias-for-git-users-fowfliwo)

## Tests

Here are some suggested E2E tests for the page based on your provided details:

### General Tests:
1. **Tab Navigation**
   - **Test:** Ensure the user can navigate between the "Drafts", "Scheduled", and "Published" tabs.
   - **Scenario:** Click on each tab and verify that the correct content is displayed.

### For the "Published" Tab:
2. **Display Published Date**
   - **Test:** Verify the correct display of "Published on" and "Last updated on" dates.
   - **Scenario:** For a post that has only been published, it should say "Published on [date]". If a post has been edited after publication, it should say "Last updated on [date]".

3. **Sorting by Published Date**
   - **Test:** Verify that posts in the "Published" tab are sorted correctly, with the most recently published or updated post appearing first. 
   - **Scenario:** Check that the order of posts reflects the latest publication or update date, with updated posts using the `updatedAt` field.
   **Note** Refer to issue #1095 by @HarshitT00 to see how the "Published posts" should be sorted. 
[...]

PR #1209, opened 2024-11-02
Add e2e my-posts page tests for delete interactions 

patrickhladun:
# ✨ Codu Pull Request 💻

add more tests to https://github.com/codu-code/codu/issues/1169
## Pull Request details

- Implement tests to ensure the delete modal can be closed with both 'Cancel' and 'Close' buttons.
- Add additional articles in setup.ts to enhance testing capabilities
- Implement createArticle function to add an article so it can be deleted in test.
- Add a test to verify the deletion of a published article through the delete modal.
- Introduce utility functions `openTab` and `openDeleteModal` to streamline modal interactions in tests.

## Any Breaking changes

-  'None'

## Associated Screenshots

-  'None'

'Hello'


e2e/my-posts.spec.ts
```diff
@@ -1,6 +1,27 @@
+import type { Page } from "@playwright/test";
 import test, { expect } from "@playwright/test";
 import { articleExcerpt, loggedInAsUserOne } from "./utils";
 
+async function openPublishedTab(page: Page) {
+  await page.goto("http://localhost:3000/my-posts");
+  await page.getByRole("link", { name: "Published" }).click();
+  await expect(page).toHaveURL(/\/my-posts\?tab=published$/);
```
pkspyder007: this might create flakiness if there is a delay in clicking the button and the url being updated with the published tab query. 

we can add something like 
`await page.waitForURL('url pattern');`
then assert the url

e2e/my-posts.spec.ts
```diff
@@ -1,6 +1,27 @@
...
+}
+
+async function openDeleteModal(page: Page) {
+  await page.getByRole("button", { name: "Options" }).click();
+  const optionsDiv = page.locator(
+    "div[aria-labelledby='headlessui-menu-button-:r5:']",
```
pkspyder007: nitpick: can we do a little readable locator instead of this headless ui dependent one

patrickhladun: That is updated @pkspyder007 and looks much cleaner, thanks for the tip.

e2e/my-posts.spec.ts
patrickhladun: Thank you for this, @pkspyder007; this is added, too.

e2e/setup.ts
```diff
@@ -20,70 +21,164 @@ export const setup = async () => {
...
-          slug: "e2e-test-slug-draft",
-          likes: 10,
-          readTimeMins: 3,
-          title: "Draft Article",
-          body: articleContent,
-          userId: authorId,
-        })
-        .onConflictDoNothing()
-        .returning(),
+    const articles: Article[] = [
```
NiallJoeMaher, line 38: Check out the seed file in how we can generate data consistently. Will be a good follow up pr

NiallJoeMaher (approved):

PR #1216, opened 2024-11-12
Feature/complete onboarding form

### LINKED ISSUE #1128: Add onboarding complete check on login
## Context

We are introducing a new onboarding form to collect user metrics.
All users will be required to input this data.

## Expected Behavior

A new field should be added to the user table to track whether a user has completed the onboarding process.

Task:
- Add a new Date field in the user table with the name "onboardingCompletedAt".
- This field should be set with the current timestamp when the user finishes the final step of the onboarding process.
- On login, check if this field is null. If it is, redirect the user to the onboarding form and prevent them from accessing the site until the form is completed.

- Additionally, we would like to add a short message, explaining to existing users why we want to gather this information. This will not be necessary for new users. We should perform a check on the "createdAt" timestamp on the user table and only display the modal if the date is earlier than the introduction of the onboarding form.
- For now, just add a placeholder for the message, we will add it later.

## Additional info

The onboarding form is found at

app>(app)>alpha>additional-details

Any questions feel free to ask.

Connected issues
#1003
#1004 
#1005 
#1006 
#1008 
#1009

John-Paul-Larkin:
# ✨ Codu Pull Request 💻

Fixes #(issue)

## Pull Request details

- This PR completes the onboarding form

- Add a ‘years of experience’ select question when the user selects ‘Working professional’
- Create migration to add two fields to the user table
    yearsOfExperience text field and onboardingComplete timestamp field

- Adds a check for completion and redirect on Login
- Perform isOnboardingComplete database check in the session callback 
- Add isOnboardingComplete boolean to the session for onboarding status tracking.
- Add component within AuthProvider to verify if the user has completed onboarding and redirect if not.

## Attempts to handle the onboarding redirect - Open to suggestions
- Middleware: Attempted to use middleware, but it can’t access the session or database.
- JWT Strategy: Tried adding an isOnboardingComplete field to the JWT, but strategy: "jwt" doesn’t integrate well with our session setup.
- Root Layout Check: Added a server-side check in the RootLayout, but this led to repeated redirects.

- In the end I added the check client side to the AuthProvider. Not ideal, but it works.  

## Any Breaking changes

- I altered the E2E setup.ts to prevent breaking tests.

## [Optional] What gif best describes this PR or how it makes you feel
SpongeBob


context/AuthProvider.tsx
```diff
@@ -1,10 +1,31 @@
...
+      <OnboardingCheck>{children}</OnboardingCheck>
+    </SessionProvider>
+  );
+}
+
+function OnboardingCheck({ children }: { children: React.ReactNode }) {
+  const { data: session, status } = useSession();
+  const router = useRouter();
+
+  console.log("-------------");
```
NiallJoeMaher: Log

server/db/schema.ts
```diff
@@ -217,6 +217,11 @@ export const user = pgTable(
     levelOfStudy: text("levelOfStudy"),
     course: text("course"),
     role: role("role").default("USER").notNull(),
+    yearsOfExperience: text("yearsOfExperience"),
```
NiallJoeMaher: We should probably use ENUMs here. 
```
CREATE TYPE experience_range AS ENUM ('0-1', '1-3', '3-5', '5-8', '12+');
```

NiallJoeMaher (changes requested):

context/AuthProvider.tsx
CarolinaCobo: @John-Paul-Larkin will you remove this log and address the coderabbitai comments too so I can merge? :)

PR #1329, opened 2026-03-09
Fix phantom notifications badge showing count but empty page

NiallJoeMaher:
## Summary
- Notification count badge showed unread notifications, but the notifications page rendered nothing
- Root cause: `getCount` counted all notifications, but the UI only renders type 0 (comment) and type 1 (reply) with non-null `post` and `notifier` — orphaned notifications from the schema migration were counted but invisible
- Added server-side filtering to both `getCount` and `get` queries to only include renderable notifications (type in [0,1], non-null postId and notifierId)

## Test plan
- [ ] Verify notification badge count matches visible notifications on the page
- [ ] Verify "Mark all as read" clears all notifications correctly
- [ ] Verify new comment/reply notifications still appear and can be dismissed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/server/api/router/notification.ts
+++ b/server/api/router/notification.ts
@@ -5,7 +5,7 @@ import {
   DeleteNotificationSchema,
 } from "../../../schema/notification";
 import { notification } from "@/server/db/schema";
-import { count, desc, eq } from "drizzle-orm";
+import { and, count, desc, eq, inArray, isNotNull } from "drizzle-orm";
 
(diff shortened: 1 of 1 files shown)


NiallJoeMaher:
Re-triggering CI — the March 9 failure pre-dates recent e2e reliability improvements on develop; testing with fresh run.

NiallJoeMaher:
Reopened to run fresh CI.

PR #1342, opened 2026-08-12
fix(admin): move the moderation preview off the public reader routes

NiallJoeMaher:
Follow-up to #1340. A high-effort review of that PR (run after it had already merged) found the *approach* was the problem, not just details of it.

## What #1340 got wrong

It let admins resolve `in_review`/`rejected` posts at their **public** URLs. Three consequences:

1. **The public reader has engagement controls.** Preview rendered vote, bookmark and comment UI on a post that may be about to be rejected — and `post.vote` has no status guard, so one misclick writes a vote and author reputation points onto content the moderator is declining.
2. **Admins stopped seeing the site as readers do.** Checking whether a takedown took effect returned 200 for an admin where everyone else gets 404.
3. **The rejected-post banner became untrue.** It still promises authors "This post is not visible to anyone else."

Plus two on the link path:

4. **Link submissions were never actually readable.** Preview went straight off-site, so the Codú-side title/excerpt/body — where a spammer would put the payload — was never shown. `gatePublish` routes links to `in_review` precisely on spam signals, so this was the case the feature existed for.
5. **That off-site href skipped `safeExternalHref` and `rel`.** An `externalUrl` that never passed `httpUrl()` validation (feed ingestion, legacy rows) would run as a `javascript:` URL inside the authenticated admin session, and the destination under review received `https://www.codu.co/admin/moderation` as its referrer.

## What this does

Preview moves to `/admin/moderation/preview/{id}` — read-only, no engagement controls, inside the existing admin gate. It renders title, author, flag reason, excerpt, tags and body, and for a link submission also the destination (through `safeExternalHref`, with `rel="noopener noreferrer nofollow"`).

The public reader routes and `postVisibilityFilter` revert to what they were, so the only person who can reach an unpublished post there is its author, as before.

Body rendering moves into a shared `PostBody` used by both the preview and the public reader, so the two cannot drift.

Keying by post id also means Preview is available for every queued post, including one whose author has not set a username — #1340 silently rendered no button in that case.

## Verified locally

- Admin at a public `in_review` URL → 404 again (both `/{username}/{slug}` and `/d/{slug}`); published posts still 200.
- Preview renders for article, question and link posts; the link preview shows the member's body *and* the guarded destination.
- Public reader still renders bodies correctly through the extracted component.
- `npm run lint`, `npm run prettier`, `npm run test:unit`, `npm run build`.


NiallJoeMaher:
Superseded by #1344, which consolidates this with #1343 and adds the email-origin fix.

PR #1343, opened 2026-08-12
fix(discussion): drop the frozen sort snapshot, serialise votes per comment

NiallJoeMaher:
Follow-up to #1341. A high-effort review of that PR (run after it had already merged) found the score-freezing was both more than the fix needed and wrong on its own terms.

## What #1341 got wrong

- **Ordering could contradict the counts next to it.** Displayed scores still updated on a refetch (window refocus, posting a comment) while the frozen ordering did not — so a thread could render a 42-point comment sitting *below* a 3-point one.
- **The escape hatch never fired.** The code comment promised "re-pick the sort to get the current ranking", but `setSortOrder("top")` when already on `top` is a no-op, React bails out, and the snapshot is never cleared.
- **It silently changed default ordering.** The added newest-first tiebreak made the Top tab identical to New on the common all-zero thread; the previous code relied on `Array#sort` stability to keep the server's order.
- **It removed the in-flight guard without replacing it.** Overlapping requests for the same comment could land in either order, leaving the stored vote disagreeing with the UI — and every rapid tap burned a request against the 100-per-5-minutes limit.
- **One failed vote wiped every optimistic vote.** `voteResetKey` was global, so an error remounted all controls and discarded the state of votes on other comments, including ones still in flight.

## What this does

**Not refetching after a successful vote is the whole fix on its own.** The new count lives in `VoteControl`; `discussions` is untouched, so nothing re-ranks under the reader. Ordering goes back to `b.score - a.score` derived from the data on screen, so it can never contradict what is displayed — and the stable-sort behaviour for ties is preserved.

**Votes are serialised per comment**, newest click replacing any queued one. One request in flight per comment, so writes cannot land out of order.

**The resync remount is per comment**, so one failure no longer discards other comments' optimistic state. Server messages (rate limiting) now reach the toast instead of a generic string.

Net: 55 insertions, 75 deletions — this is mostly deleting the machinery #1341 added.

## Verified locally

- Vote → order unchanged, count updates.
- 8 clicks dispatched back-to-back → **2** requests (was 8), and the stored vote + count match the UI exactly (`up`, `upvotes_count = 1`).
- 6 clicks with small gaps → serialised, DB and UI agree.
- `npm run lint`, `npm run prettier`, `npm run test:unit`, `npm run build`.

### DIFF
--- a/components/Discussion/DiscussionArea.tsx
+++ b/components/Discussion/DiscussionArea.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import React, { useMemo, useRef, useState } from "react";
+import React, { useRef, useState } from "react";
 import {
   Menu,
   MenuButton,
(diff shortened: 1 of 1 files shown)


NiallJoeMaher:
Superseded by #1344, which consolidates this with #1342 and adds the email-origin fix.

PR #1345, opened 2026-08-12
[do not merge] e2e baseline at 2ac41e9d

NiallJoeMaher:
Throwaway draft to get an E2E baseline at the commit *before* #1340/#1341/#1344. The develop runs before that point were all cancelled by concurrency, so there is no clean signal to compare against. Empty commit; will be closed as soon as the run reports.


NiallJoeMaher:
Baseline captured: 2ac41e9d (before #1340/#1341/#1344) fails with exactly the same 9 specs as develop, so the e2e failures predate all of that work. Fix is in #1344.

PR #1346, opened 2026-08-13
[do not merge] e2e harness change on develop, no app changes

NiallJoeMaher:
Isolates the e2e harness change (production build + AUTH_TRUST_HOST) from the app changes in #1344, so the two new failures there can be attributed to one or the other. Will close as soon as it reports.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,8 @@
   "scripts": {
     "dev": "next dev",
     "dev:e2e": "ENV=E2E next dev",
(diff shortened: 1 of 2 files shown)


NiallJoeMaher:
Isolation result captured: with only the harness change on develop, articles.spec.ts:66 and notifications.spec.ts:192 both still fail — so neither is caused by the app changes in #1344. Recorded there.

